PR #47, opened 2025-10-18
CPU Usage Percentages Fix

danfq:
This fixes the CPU Usage Percentages.
They are now Core-accurate, meaning they won't go over 100% (real utilization, normalized across all CPU Cores).


Ebullioscopic (changes requested):
image

The CPU processes usage is still giving crazy numbers

danfq:
image

My previous testing was on my Ryzentosh, and everything was fine.
Strangely, I'm actually seeing different results, here.

On my M3, the percentages seem to be "stuck" at 0.0%, for the most part.
I'll work on this as soon as I get back home, as I need to go to a business dinner, now.

I will be bringing my M3, though, so if you need anything done quick, shoot me a DM on Insta (I sent you one yesterday).

PR #48, opened 2025-10-18
Feature/hud

Ebullioscopic:
# Fix - Works without the OSDUIHelper Disabler 

### -> Also works on macOS Tahoe

Steps to allow privacy accessibility without conflicts in multiple builds:

1. Copy `DynamicIsland.app` to `Applications` Folder

2. `Product` > `Scheme` > `Edit Scheme` > Change `Executable` to `Other` and select the `DynamicIsland.app` in `Applications` folder


SuperAnthony2317:
but after I copied the DynamicIsland, and changed the running app to the app in the Application folder, do i also change the other exutables like profiles? And do i just have to use the app form the launchpad and not directly in xcode?
Screenshot 2025-10-21 at 17 44 05
Screenshot 2025-10-21 at 17 44 10

PR #54, opened 2025-10-20
implement download and musicvisualizer feature

macanoclothing-blip:


Ebullioscopic:
When some file is being downloaded - I cannot access the island during that period

https://github.com/user-attachments/assets/64a42dd5-631d-4aaa-984b-fb6e03e29f1a

image

Also, there are alignment issues with activity - the height isn't matching the generic menubar height

Similarly, when I cancel the download of the file - the progress gets stuck - and I cannot use the notch anymore

Please fix these issues

PR #73, opened 2025-10-28
Fix Bluetooth Headphone Volume Imbalance when HUD replacement enabled.

### LINKED ISSUE #70: Enabling built in system hud only reduces/increases volume in left ear when using bluetooth headphones

Sailokeshg:
## Fix Bluetooth Headphone Volume Imbalance

When using the built-in system HUD for volume control with Bluetooth headphones, sound was only coming from the left ear. This was happening because the volume controller was trying to adjust individual audio channels (left/right) instead of using the master volume control that Bluetooth devices expect.

**What I changed:** Added Bluetooth device detection to the `SystemVolumeController` and modified it to prioritize the master volume element for Bluetooth devices, while keeping the existing behavior for wired audio devices.

**Why this works:** Bluetooth audio devices handle stereo balance differently than traditional audio interfaces. By focusing on the master volume control instead of individual channels, we ensure both ears get equal volume adjustments. The fix is backwards compatible and won't affect other audio devices.

**Testing:** Verified with Bluetooth headphones (Airpods Pro 2 )- now both ears play balanced audio when using volume keys with the system HUD enabled. Wired devices and other audio outputs continue to work as before.

This should resolve the stereo imbalance issue users were experiencing with Bluetooth audio devices.

Fixes #70

### DIFF
--- a/DynamicIsland/managers/SystemMediaControllers.swift
+++ b/DynamicIsland/managers/SystemMediaControllers.swift
@@ -418,17 +418,50 @@ final class SystemVolumeController {
     }
 
     private func volumeElements() -> [AudioObjectPropertyElement] {
-        candidateElements.filter { element in
+        // For Bluetooth devices, prioritize master element to avoid stereo balance issues
+        if isBluetoothDevice() {
+            if candidateElements.contains(kAudioObjectPropertyElementMaster) {
+                var address = makeAddress(selector: kAudioDevicePropertyVolumeScalar, element: kAudioObjectPropertyElementMaster)
+                if propertyExists(deviceID: currentDeviceID, address: &address) {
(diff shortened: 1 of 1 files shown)


Sailokeshg:
So, we are not resolving this issue now? @Ebullioscopic

Ebullioscopic:
I tried to replicate this issue with my Airpods Pro 2, OnePlus Buds, JBL Headphones - on M4 Pro (Sequoia and Tahoe), M3 Pro, M2 Air and M1 Air - wasn't able to replicate it anywhere with the latest version (but started reflecting when I applied your changes) - I can reopen it if you are still facing the issue with the latest commit - lemme know what your views are, @Sailokeshg

yl25946:
Fixed on newest version.

PR #200, opened 2025-12-25
Several Changes Reuploaded

Ibrty:
Fixed the Alignment of the Weather and Focus Lock Screen Widgets Changed the Lock screen behaviour so the focus mode now renders below the weather widgets. Added the "Reduce Interruptions" focus mode and its corresponding icon and color. Added the option to toggle "Only show focus icon" which hides the name of the active focus mode on the lock screen Added Support for Beats headphones to the lock screen Bluetooth battery widget Changed the Bluetooth battery widget behaviour to allow it to display all connected Bluetooth devices, rather than whichever device was the most recently connected

### DIFF
--- a/DynamicIsland/AudioRouteManager.swift
+++ b/DynamicIsland/AudioRouteManager.swift
@@ -1,0 +1,245 @@
+import Combine
+import CoreAudio
+import Foundation
+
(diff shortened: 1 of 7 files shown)


Ibrty:
Closing to fix issues

PR #255, opened 2026-01-21
Fix: Prevent '[]' from displaying when no lyrics are available

### LINKED ISSUE #254: Issue: Lyrics display shows "[]" when no lyrics are available
# Lyrics display shows "[]" when no lyrics are available

When the lyrics API returns an empty response (e.g., an empty JSON array []), the app displays the literal string "[]" in the lyrics UI instead of showing nothing or hiding the lyrics section.

Image

### Steps to Reproduce

1. Play a track that has no lyrics available in the LRCLIB API
2. Observe that the string "[]" is displayed instead of an empty state

### Expected Behavior

When no lyrics are available, the lyrics section should:
Show nothing (empty/hidden state)
Not display any placeholder text like "[]"

### Actual Behavior

The app displays the literal string "[]" when the API returns an empty JSON array response.

### Root Cause

In `MusicManager.swift`, the `fetchLyricsFromAPI` method's fallback logic treats the API response as plain text when JSON parsing fails or when the response is the literal string "[]". This causes the string to be displayed as lyrics text instead of being recognized as "no lyrics available".

### Environment
macOS version: Tahoe (version 26)
App version: 1.2.2-beta (latest)

### Proposed Solution
Add a check to detect when the trimmed response string is "[]" and return an empty array instead of treating it as lyrics text:

7hourspg:
### Problem
When the lyrics API returns an empty response, the app shows the literal string "[]" in the lyrics UI.
### Solution
Added a check to detect empty JSON array responses and return an empty array instead of treating "[]" as lyrics text.
### Changes
Added validation in `fetchLyricsFromAPI` to detect "[]" responses
Return empty array when response is "[]" or empty

After fix (no lyrics available):
Screenshot 2026-01-22 at 3 25 04 AM

### Testing
[x] Tested with tracks that have no lyrics
[x] Verified empty state shows correctly
[x] Code builds without errors
Fixes #254

### DIFF
--- a/DynamicIsland/managers/MusicManager.swift
+++ b/DynamicIsland/managers/MusicManager.swift
@@ -708,14 +708,27 @@ class MusicManager: ObservableObject {
                 }
             } else {
                 // Fallback: try to decode as UTF8 and handle as LRC or plain text
-                if let lrcString = String(data: data, encoding: .utf8), !lrcString.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty {
-                    // If it contains a syncedLyrics key in an object, try that
-                    if let json = try? JSONSerialization.jsonObject(with: data, options: []) as? [String: Any],
-                       let synced = json["syncedLyrics"] as? String {
-                        return parseLRC(synced)
+                if let lrcString = String(data: data, encoding: .utf8) {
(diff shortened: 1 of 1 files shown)


danfq (approved):

danfq:
lgtm

@Ebullioscopic merge when you can please

PR #260, opened 2026-01-22
Update Feedback for Volume Change

### LINKED ISSUE #250: Play a sound when adjusting volume with Volume HUD enabled
It'd be nice if the Volume HUD could play a sound when the volume is adjusted.

When Atoll overrides the built in Volume HUD, it inadvertently disables the built in volume click (if you have it enabled) which I find useful for guaging the current volume when nothing is playing.

Ebullioscopic:
Fixes #250


Ebullioscopic:
image

Issue #332, opened 2026-03-09
Wrong Focus label for Sleep


SuperAnthony2317:
<img width="1470" height="956" alt="Image" src="https://github.com/user-attachments/assets/3b58ef8f-21dc-4c75-ab45-a6f3f9069481" />
*Other Focus like "Gaming" or  Personal" works

Ebullioscopic:
<img width="566" height="101" alt="Image" src="https://github.com/user-attachments/assets/c8586fe5-64c9-4722-96f8-5554978c3b53" />

@SuperAnthony2317 can you try opening Atoll's `Settings` > Menu Bar `Focus` > `Use DevTools` and restart Atoll once

If it is still not functioning properly, lemme know your macOS Tahoe version or if you're running `beta`

Ebullioscopic:
Thanks for clarifying @SuperAnthony2317 , I'm running on macOS 26.3.1 stable and hence not able to replicate this issue. I'll try to find a fix to why it is broken in the `beta`

StudioKeys:
@Ebullioscopic close this since its fixed since the last release

PR #334, opened 2026-03-12
Atoll: Fix Sleep Mode Icon Detections on Atoll

StudioKeys:
image


StudioKeys:
> #332 @Ebullioscopic this is now covered in this PR, I have seen that if you have a focus mode enabled prior to launching the app, it doesn't show the same on the app

Both are now covered in this PR now, and verified on my M4 Air running macOS Tahoe 26.3.1

PR #414, opened 2026-04-13
Added Amazon Music Support

sharma2007:
Adds Amazon Music as a selectable media backend so the notch can show title, artist, artwork, and playback state when Amazon Music (`com.amazon.music`) is the active Now Playing source, with transport controls routed through the same Media Remote path as other bundled controllers.

New `AmazonMusicController` — Subscribes to the existing `mediaremote-adapter` JSON stream, applies updates only when the stream identifies `com.amazon.music` as the source, and idles the UI when another app owns Now Playing so stale Amazon metadata is not shown.
`MediaControllerType` — New Amazon Music case, wired in `MusicManager` (controller factory, accent color).
System integration — `com.amazon.music` included where other music bundle IDs are listed (e.g. app/audio routing helpers) so behavior stays consistent with Apple Music, Spotify, and YouTube Music.
Settings & onboarding — Short copy explaining that controls follow system Now Playing when Amazon Music is the active source.
Localization — String catalog entries for the new option and descriptions.

Note: Requires Amazon Music to be the relevant Now Playing participant; behavior is aligned with macOS Now Playing semantics described in-app.

### DIFF
--- a/DynamicIsland/DynamicIslandApp.swift
+++ b/DynamicIsland/DynamicIslandApp.swift
@@ -184,6 +184,7 @@ class AppDelegate: NSObject, NSApplicationDelegate {
         let targetBundleIDs = [
             "com.apple.Music",
             "com.spotify.client",
+            "com.amazon.music",
             "com.apple.Safari",
             "com.tidal.desktop",
             "tv.plex.plexamp",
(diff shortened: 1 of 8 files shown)


sharma2007:
Hey, fixed the localization file. Sorry for the inconvenience :)

Issue #420, opened 2026-04-16
[Bug] Terminal tab: keyboard shortcut opens then immediately closes; sticky mode workaround traps session (outside click doesn’t dismiss)


sharma2007:
Opening the in-notch Terminal with the global Toggle Terminal Tab shortcut often collapses the notch almost immediately, even from a fully closed notch and with no double keypress. Turning on Terminal sticky mode avoids the auto-close but introduces a second problem: clicking outside no longer dismisses the terminal, so the only practical way to close is to toggle sticky mode off in Settings, which is cumbersome and easy to misread as “the app is stuck.”

**Steps to reproduce:** In the attached video.

**Expected behavior:**
- A single shortcut press from a closed notch should open the terminal tab and keep it open until the user explicitly dismisses it (shortcut again, close control, or documented dismiss gesture), consistent with other tabs opened without hover.
- With sticky mode enabled, click outside should still close the terminal/notch (or another equally discoverable dismiss path), without forcing a round-trip through Settings.

**Likely Cause:**
- In `ContentView`, a change in `dynamicNotchSize` (inevitable when switching to Terminal’s tall layout) triggers `shouldRecheckHover`, whose handler closes the notch when `notchState == .open`, `!isHovering`, and `!shouldPreventAutoClose()`.
-  The global “click outside to close” handler is only installed in `handleHover(false)` after the pointer has left the notch (`ContentView`). Opening the terminal only via keyboard may never run that hover path, so no outside-click monitor is registered.

**Environment:**
- macOS: Sequoia 15.7.3
- Atoll Version: v2.1.0 Seychelles

https://github.com/user-attachments/assets/89b59854-0d76-4b7d-8605-977d13ff7564

Ebullioscopic:
Hi @sharma2007 , yeah this looks like a fixable bug, if you're willing you can work on it. I haven't got a chance to review your previous PR yet, will review and merge it in a couple days.

sharma2007:
Hey @Ebullioscopic, I've fixed this and created a PR.

PR #421, opened 2026-04-16
Fix terminal keyboard open + sticky outside-click

sharma2007:
**Root causes**:
- **Snap shut:** `dynamicNotchSize` updates (e.g. switching to Terminal’s tall layout) toggled `shouldRecheckHover`, which ran the “popover-style” path that calls `vm.close()` when the notch is open, not hovering, and auto-close is allowed. Keyboard opens never set `isHovering`, so that path fired right after open.
- **Sticky + outside click:** The global outside-click monitor was only installed from `handleHover(false)` after the pointer had left the notch. Keyboard-only opens never ran that hover lifecycle, so the monitor was often never installed.
- **Timer:** `toggleNotchOpen` schedules a 3s close; `toggleTerminalTab` did not clear it when focusing Terminal.

**Changes:**
- `ContentView.swift`: Removed `onChange(of: dynamicNotchSize)` that toggled `shouldRecheckHover`.
- `ContentView.swift`: Added `syncStickyTerminalOutsideClickMonitor()` and `@Default(.terminalStickyMode)`, call sync from `onChange(of: coordinator.currentView)` and `onChange(of: terminalStickyMode)` so the sticky dismiss monitor is installed whenever Terminal + sticky + open apply, and removed when they don’t.
- `DynamicIslandApp.swift`: In `toggleTerminalTab`, cancel and clear `closeNotchWorkItem` when opening from closed or switching another tab to Terminal.

### DIFF
--- a/DynamicIsland/ContentView.swift
+++ b/DynamicIsland/ContentView.swift
@@ -82,6 +82,7 @@ struct ContentView: View {
     @Default(.showStandardMediaControls) var showStandardMediaControls
     @Default(.externalDisplayStyle) var externalDisplayStyle
     @Default(.hideNonNotchUntilHover) var hideNonNotchUntilHover
+    @Default(.terminalStickyMode) var terminalStickyMode
     
     // Dynamic sizing based on view type and graph count with smooth transitions
     var dynamicNotchSize: CGSize {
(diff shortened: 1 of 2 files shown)


sharma2007:
issue #420

Issue #371, opened 2026-03-29
[Bug] Minor terminal bugs


brucemakes012:
Hey, I've experienced some small bugs using the Terminal extension (is extension the right term?). 

1. The shortcut doesn't appear to work. I've assigned it to ⌘ + ⌥ + T, but when pressing this nothing happens
2. Not sure if this is the intended effect or not, but when lowering the opacity of the terminal window, instead of becoming more transparent, it simply becomes darker. I'd love if it can achieve actual transparency, like I have setup for my own Terminal here:

<img width="417" height="120" alt="Image" src="https://github.com/user-attachments/assets/e7020ec2-233c-46b9-8b83-9c8fb1ebb9b3" />

3. The terminal cursor doesn't update to match what's in the settings. You can see here that I've selected underline, but the cursor is always "block (blinking)" no matter which option I pick

<img width="567" height="472" alt="Image" src="https://github.com/user-attachments/assets/b1de8bfa-16f4-41a7-8d17-8352f7666255" />

I'm on Atoll 2.1.0 and Tahoe 26.4. Let me know if there's anything else I can provide to help, thanks!

sharma2007:
Point 1: Fixed in PR #421 
Point 3: Fixed in PR #423 

As for point 2, currently the terminal opacity slider changes the opacity of the text. @Ebullioscopic is the intended behaviour how @brucemakes012 showed in his screenshot? If yes, I can fix that and add that to my PR #423.

Ebullioscopic:
Yes @sharma2007 this is the intended behaviour for terminal opacity, do make sure that it does not change the opacity of all the tabs, only the terminal tab

brucemakes012:
@sharma2007 thanks for working on this! I just had a thought about the opacity thing. Since the other tabs need to remain fully opaque, maybe having the area marked in red be a gradient from the terminal opacity (less than 100%) to fully opaque at the top would be a good way to do it. I'm not a developer so I'm not sure how feasible this suggestion is, but I thought it was a visually cool way to have a translucent terminal tab while the other tabs are opaque and wanted to share. Hope this helps!

<img width="634" height="441" alt="Image" src="https://github.com/user-attachments/assets/72fb427a-59d2-4ac7-8ea6-bc5631b72177" />

sharma2007:
Hey @brucemakes012, thanks for the idea! This is what I have so far:

[...]

PR #423, opened 2026-04-20
[Bug Fix] Terminal Bug(s)

sharma2007:
Fixes point 3 of issue #371. By using `setCursorStyle` so the delegate updates `CaretView`(`options` alone does not)

### DIFF
--- a/DynamicIsland/managers/TerminalManager.swift
+++ b/DynamicIsland/managers/TerminalManager.swift
@@ -218,6 +218,9 @@ class TerminalManager: ObservableObject {
             ?? .blinkBlock
         view.getTerminal().setCursorStyle(cursorStyle.swiftTermStyle)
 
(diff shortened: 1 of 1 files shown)


sharma2007:
Additionally currently in scenarios like opening the terminal via shortcut or having terminal + settings open side by side and changing something in settings, changes the cursor to a block outline. The renderer is switching to the “no keyboard focus” caret appearance (as in the picture). @Ebullioscopic is that the intended behaviour?
image

Ebullioscopic:
I mean, it is not intended for sure, lemme know if you can address that in the same PR here, I'll merge it alongside

sharma2007:
@Ebullioscopic Made the required change. Let me know the intended behaviour for terminal opacity and I can implement that to this PR as well.

Ebullioscopic:
@sharma2007 I'll merge this PR as it is right now, I think you can create a new PR for the terminal transparency

PR #430, opened 2026-04-24
Terminal opacity fix

sharma2007:
Old behavior mixed multiple opacity paths (`nativeBackgroundColor` alpha, layer background state, view composition). Result: inconsistent translucency, occasional wrong visual blending, confusing UX around what opacity slider controls.

Fixes:
- Compute terminal background color with slider alpha in `nativeBackgroundColor`.
- Force terminal layer to transparent (`layer.backgroundColor = .clear`, non-opaque) after settings application paths so underlay blur can show through correctly.
- Trigger redraw after translucency sync for live updates.
- Keep container/effect/terminal frames aligned to same bounds for predictable compositing.
- Keep settings text explicit: opacity controls terminal backdrop; text remains fully opaque.


sharma2007:
Fixed a couple of bugs, next is the text padding.

PR #441, opened 2026-04-30
fix(managers): force-respawn OSDUIHelper on kickstart so native HUD stays suppressed at first launch

nightah:
Pass `-k` to `launchctl kickstart` so launchd kills any running OSDUIHelper instance and respawns it cleanly before the subsequent SIGSTOP. Without `-k`, kickstart is a no-op when the service is already running, and the SIGSTOP can land on a lingering instance that macOS then replaces when a media key is pressed, leaving the native HUD visible on first launch until the user toggled the setting off and on again. With `-k`, the stop reliably applies to the freshly spawned process and the native HUD is suppressed immediately.


Ebullioscopic:
Hi @nightah , I believe the OSDUIHelper killing way is an old implementation, current one is basically supressing the HUDs by directly passing the data via the accessibility permissions. Let me know if I'm getting smth wrong

nightah:
@Ebullioscopic, you're right, and on further inspection and testing, I think the actual issue is that the `MediaKeyInterceptor` sometimes isn't launching successfully.

I often have multiple HUDs appearing like below:

image

I find there are two ways to resolve this:

1. Go into the settings and change the HUD (this seems to re-register the interceptor).
2. Restart Atoll

I'm not certain what is causing it not to work. I'll try to reproduce the issue and validate/update this PR.

StudioKeys:
Marking it as WIP to avoid accidental merges

PR #442, opened 2026-05-01
Add configurable step sizes for volume and brightness keys

### LINKED ISSUE #171: Custom HUD steps
Being able to control the step sizes (normal & small) for volume/brightness would be awesome as it's not build into MacOS.  MediaMate supports this and I was wondering if it would be possible to add to Atoll?

lugoues:
## Summary

Adds configurable step sizes for hardware volume and brightness keys. Previously hardcoded at 6% and 2% (Shift+Option fine), which many users find too coarse.

Closes #171 

## Notes/Issues/Questions
- The "DCC app integration" section is pinned to the bottom and doesn't scroll  (see screenshot).
- No translations yet - how should I handle this?
- I didn't add a toggle for the entire block (see screenshot) and instead just default to the system defaults. Do we want a toggle for the entire "Step size" block?

## Changes

- Added new `Defaults` keys: `volumeStepPercent`, `volumeFineStepPercent`, `brightnessStepPercent`, `brightnessFineStepPercent` (Int, range 1-25).
- `SystemChangesObserver` now reads these as computed properties for immediate application on keypress.
- Updated Settings UI with a new "Step size" section.
- Defaults of 6 and 2 preserve existing behavior.

## Testing

- [x] Defaults produce existing behavior (6%, 2%)
- [x] Setting volume step to 1 works
- [x] Shift+Option uses fine step
- [x] Brightness behaves identically
- [x] Mute key still works
- [x] Settings UI updates take effect immediately

## Screenshots

image

### DIFF
--- a/DynamicIsland/Localizable.xcstrings
+++ b/DynamicIsland/Localizable.xcstrings
@@ -8017,6 +8017,9 @@
         }
       }
     },
+    "Brightness fine step" : {
+
+    },
     "Brightness HUD" : {
       "localizations" : {
         "fr" : {
@@ -8080,6 +8083,9 @@
           }
         }
       }
+    },
+    "Brightness step" : {
+
     },
     "Built-in" : {
       "comment" : "A label indicating that an animation is a built-in one.",
@@ -14840,6 +14846,9 @@
         }
       }
     },
+    "Disabled while external display brightness integration is active." : {
+
+    },
     "Disabled while external display integration is active — brightness keys are handled by the external app." : {
       "localizations" : {
         "ko" : {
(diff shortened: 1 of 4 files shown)


Ebullioscopic:
Hi @lugoues , thanks for the PR, I think you can make the External DDC app integration a collapsible pinned to the bottom. You can create the translations in a new PR, and ig the current implementation of step size is good enough

lugoues:
@Ebullioscopic Is this what you were thinking?
image
image

lugoues:
@Ebullioscopic checking in to see if you have thoughts on the above.

Ebullioscopic:
Yeah @lugoues the changes look good

I'll try to merge it soon, just a bit caught up with other stuff

lugoues:
No rush, I’ll update the PR with the UI changes over the weekend.
Thanks!

lugoues:
@Ebullioscopic 
I've split the step and ddc config into a new tab. Having ddc at the bottom of "controls" took a lot of space away from the rest of the page and step settings would have been duplicated across each OSD/HUD tab.

We now have "Controls" which configure the controls, and "Display Overlay" to configure HUD/OSD.

image

PR #479, opened 2026-05-29
Fix brightness HUD reverting to native macOS OSD on Tahoe

### LINKED ISSUE #478: Brightness HUD reverts to native macOS OSD on Tahoe — CGEvent tap never re-enabled after timeout
## Bug

On macOS Tahoe (26.x), the Dynamic Island brightness HUD stops appearing after a short period of use. The native macOS brightness slider takes over permanently until the app is restarted. Volume HUD continues to work.

**Atoll version:** 2.2.0 (Fiji)  
**macOS:** Tahoe 26.5 (Darwin 25.5.0)  
**Hardware:** Apple Silicon

## Root Cause

`MediaKeyInterceptor.handleEvent` does not handle `CGEventType.tapDisabledByTimeout` or `.tapDisabledByUserInput`. When macOS detects the event tap callback is taking too long, it disables the tap automatically. The brightness callback path is slower than volume because `SystemBrightnessController.adjust(by:)` calls `syncWithSystemBrightnessIfNeeded()` synchronously — this reads the display brightness through CoreBrightness/DisplayServices/IODisplay, which can be slow enough on Tahoe to trigger the timeout. Once disabled, the tap is never re-enabled, so all subsequent brightness (and volume) key events pass through to macOS, which shows the native OSD.

**Relevant code in `MediaKeyInterceptor.swift`:**

```swift
private func handleEvent(cgEvent: CGEvent, type: CGEventType) -> Unmanaged<CGEvent>? {
    // ❌ No handling for .tapDisabledByTimeout / .tapDisabledByUserInput
    guard let systemDefinedType = systemDefinedEventType,
          type == systemDefinedType,
          let nsEvent = NSEvent(cgEvent: cgEvent),
          nsEvent.subtype.rawValue == 8 else {
        return Unmanaged.passUnretained(cgEvent)  // silently passes through the disable notification
    }
```

### 1. `MediaKeyInterceptor.swift` — Re-enable tap on system disable

Add handling at the top of `handleEvent` to catch the disable event and re-enable the tap:

```swift
private func handleEvent(cgEvent: CGEvent, type: CGEventType) -> Unmanaged<CGEvent>? {
    if type == .tapDisabledByTimeout || type == .tapDisabledByUserInput {
        if let tap = eventTap {
            CGEvent.tapEnable(tap: tap, enable: true)
            NSLog("⚠️ Media key event tap was disabled by the system (reason: %@); re-enabled",
                  type == .tapDisabledByTimeout ? "timeout" : "user input")
        }
        return Unmanaged.passUnretained(cgEvent)
    }
    // ... rest of handleEvent unchanged
```

### 2. `SystemMediaControllers.swift` — Remove slow sync from event tap callback path

The `syncWithSystemBrightnessIfNeeded()` call in `adjust(by:)` does synchronous system calls inside the event tap callback, which is what triggers the timeout in the first place. Remove it — the notification handlers and polling timer already keep `lastEmittedBrightness` in sync, and `beginBrightnessAnimation` syncs before starting.

Also add `pendingAdjustTarget` to fix rapid key presses collapsing into a single step (the async dispatch means multiple `adjust()` calls read the same stale `lastEmittedBrightness`):

```swift
private var pendingAdjustTarget: Float?

func adjust(by delta: Float) {
    markUserInitiated()
    let base = pendingAdjustTarget ?? lastEmittedBrightness
    let target = max(0, min(1, base + delta))
    pendingAdjustTarget = target
    DispatchQueue.main.async { [weak self] in
        guard let self else { return }
        self.pendingAdjustTarget = nil
        self.beginBrightnessAnimation(to: target)
    }
}
```

### 3. `SystemChangesObserver.swift` — Add missing guard to brightness handler

The volume handler has `guard volumeEnabled else { return }` at the top, but the brightness handler was missing an equivalent guard, causing unnecessary work when brightness control is disabled:

```swift
func mediaKeyInterceptor(
    _ interceptor: MediaKeyInterceptor,
    didReceiveBrightnessCommand direction: MediaKeyDirection,
    step: MediaKeyStep,
    isRepeat: Bool,
    modifiers: NSEvent.ModifierFlags
) {
    let isBacklight = modifiers.contains(.command) && keyboardBacklightEnabled
    guard isBacklight || brightnessEnabled else { return }
    // ... rest unchanged
```

## Steps to Reproduce

1. Install Atoll 2.2.0 on macOS Tahoe 26.x
2. Enable Dynamic Island brightness HUD in settings
3. Press brightness keys — Dynamic Island HUD appears initially
4. After some usage (varies, sometimes immediately), the native macOS brightness slider appears instead
5. The Dynamic Island HUD never returns until Atoll is restarted

Kolomaster68:
## Summary

- **Re-enable CGEvent tap after system timeout** — macOS auto-disables event taps when the callback takes too long; the brightness path's synchronous `CoreBrightness`/`DisplayServices` calls were slow enough on Tahoe to trigger this. The tap was never re-enabled, permanently killing all media key interception.
- **Remove slow sync from tap callback path** — moved `syncWithSystemBrightnessIfNeeded()` out of `adjust(by:)` to prevent future timeouts. Added `pendingAdjustTarget` to fix rapid key presses collapsing into a single brightness step.
- **Add missing early guard to brightness handler** — matches the volume handler's `guard volumeEnabled` pattern.

Fixes #478

## Test plan

- [ ] On macOS Tahoe, press brightness keys — Dynamic Island HUD should appear instead of native OSD
- [ ] Hold brightness key for rapid repeat — brightness should change by one step per repeat (not collapse)
- [ ] Disable brightness HUD in settings — brightness keys should pass through to native macOS handling
- [ ] Cmd+brightness (keyboard backlight) should still work when backlight control is enabled
- [ ] Volume HUD should continue working as before

### DIFF
--- a/DynamicIsland/managers/MediaKeyInterceptor.swift
+++ b/DynamicIsland/managers/MediaKeyInterceptor.swift
@@ -176,6 +176,15 @@ final class MediaKeyInterceptor {
     }
 
     private func handleEvent(cgEvent: CGEvent, type: CGEventType) -> Unmanaged<CGEvent>? {
+        if type == .tapDisabledByTimeout || type == .tapDisabledByUserInput {
+            if let tap = eventTap {
+                CGEvent.tapEnable(tap: tap, enable: true)
+                NSLog("⚠️ Media key event tap was disabled by the system (reason: %@); re-enabled",
+                      type == .tapDisabledByTimeout ? "timeout" : "user input")
+            }
+            return Unmanaged.passUnretained(cgEvent)
+        }
+
         guard let systemDefinedType = systemDefinedEventType,
               type == systemDefinedType,
               let nsEvent = NSEvent(cgEvent: cgEvent),
--- a/DynamicIsland/managers/SystemChangesObserver.swift
+++ b/DynamicIsland/managers/SystemChangesObserver.swift
@@ -184,6 +184,9 @@ final class SystemChangesObserver: MediaKeyInterceptorDelegate {
         isRepeat: Bool,
         modifiers: NSEvent.ModifierFlags
     ) {
+        let isBacklight = modifiers.contains(.command) && keyboardBacklightEnabled
+        guard isBacklight || brightnessEnabled else { return }
+
         // Elastic Limit Detection (Vertical HUD)
         if Defaults[.enableVerticalHUD] {
             let brightness = brightnessController.currentBrightness
@@ -193,12 +196,12 @@ final class SystemChangesObserver: MediaKeyInterceptorDelegate {
                 Task { @MainActor in VerticalHUDWindowManager.shared.triggerBump(direction: -1) }
             }
         }
-        
+
         let baseStep = brightnessStep(for: step)
         let delta = direction == .up ? baseStep : -baseStep
-        if modifiers.contains(.command) && keyboardBacklightEnabled {
+        if isBacklight {
             keyboardBacklightController.adjust(by: delta)
-        } else if brightnessEnabled {
+        } else {
             brightnessController.adjust(by: delta)
         }
     }
--- a/DynamicIsland/managers/SystemMediaControllers.swift
+++ b/DynamicIsland/managers/SystemMediaControllers.swift
@@ -476,6 +476,7 @@ final class SystemBrightnessController {
     private let maximumBrightnessAnimationDuration: TimeInterval = 0.3
     private let brightnessAnimationDurationScale: TimeInterval = 1.6
     private var lastEmittedBrightness: Float = 0.5
+    private var pendingAdjustTarget: Float?
     private let coreBrightnessClient = CoreBrightnessDisplayClient.shared
     private var pollTimer: Timer?
     private let pollInterval: TimeInterval = 0.15
@@ -526,12 +527,14 @@ final class SystemBrightnessController {
     }
 
     func adjust(by delta: Float) {
-        // Refresh baseline to avoid jumping if auto-brightness changed the level.
-        syncWithSystemBrightnessIfNeeded()
         markUserInitiated()
-        let target = max(0, min(1, lastEmittedBrightness + delta))
+        let base = pendingAdjustTarget ?? lastEmittedBrightness
+        let target = max(0, min(1, base + delta))
+        pendingAdjustTarget = target
         DispatchQueue.main.async { [weak self] in
-            self?.beginBrightnessAnimation(to: target)
+            guard let self else { return }
+            self.pendingAdjustTarget = nil
+            self.beginBrightnessAnimation(to: target)
         }
     }
 


StudioKeys:
This PR is not even required as i just tested the PR and even with it its broken, here is the root cause in my setup that i had which got fixed as well, putting it here

What happens: Third-party DDC integration ON → brightnessEnabled=false → interceptBrightness=false → brightness key not swallowed → passes through → native macOS OSD. By design: it hands brightness off to an external DDC app (MonitorControl/BetterDisplay) so that app drives brightness + its own OSD.

To fix: Disable DDC Integration and it should work fine as normal and things should be fine.

Thanks though for taking the time to investigate the issue and its greatly appreciated

Cc: @Ebullioscopic

StudioKeys (changes requested):
Not required, could be safely discarded

Kolomaster68:
Weird, it was only working that way through my debug version. Will try your method, there's been issues with Tahoe ever since 26.5 update.

Kolomaster68:
@StudioKeys You were right, and thanks for challenging my original assumption.

I re-tested this properly with a clean setup (single install, no debugger attached), and DDC integration turned out to be the deciding factor. It was enabled during my earlier testing. Once I disabled it, Atoll’s brightness HUD worked correctly even on the unfixed release. So the issue on my system was the DDC handoff you described, not a disabled event tap.

Given that, #478 doesn’t really stand up as reported. I wasn’t able to reproduce the tap-timeout behaviour once DDC was disabled, so this looks like a configuration issue rather than a bug. I’ll close it.

That said, re-enabling the tap on .tapDisabledByTimeout and .tapDisabledByUserInput still seems like sensible defensive programming for any CGEvent tap, regardless of this specific issue. If you think it’s worth keeping, I can trim the PR down to just that change and remove the brightness-sync modifications. Otherwise, I’m happy to close that as well.

StudioKeys:
> @StudioKeys You were right, and thanks for challenging my original assumption.
> 
> I re-tested this properly with a clean setup (single install, no debugger attached), and DDC integration turned out to be the deciding factor. It was enabled during my earlier testing. Once I disabled it, Atoll’s brightness HUD worked correctly even on the unfixed release. So the issue on my system was the DDC handoff you described, not a disabled event tap.
> 
> Given that, #478 doesn’t really stand up as reported. I wasn’t able to reproduce the tap-timeout behaviour once DDC was disabled, so this looks like a configuration issue rather than a bug. I’ll close it.
> 
> That said, re-enabling the tap on .tapDisabledByTimeout and .tapDisabledByUserInput still seems like sensible defensive programming for any CGEvent tap, regardless of this specific issue. If you think it’s worth keeping, I can trim the PR down to just that change and remove the brightness-sync modifications. Otherwise, I’m happy to close that as well.

I'm working on a DDC guard as for a defensive fix just for now which makes the UX better, but i will get to the bottom of this and make it work both-ways just to not confuse the end-user

PR #483, opened 2026-06-02
Terminal Text Padding + Side Scroll Bar Rework + Minor Close Issue Fix

sharma2007:
## Terminal inner inset added (all sides)

- Introduced `notchTerminalInnerTextInset` in `DynamicIsland/sizing/matters.swift`.
- `StableTerminalContainerView.resizeSubviews` now frames `LocalProcessTerminalView` to `bounds.insetBy(...)` instead of full bounds.
- Initial mount path in `TerminalManager.ensureTerminalView` uses the same inset frame, keeping behavior consistent between first layout and subsequent resizes.

## Inset gutter background harmonized

- Added `terminalGutterTintView` below the terminal view and above the frosted effect view.
- Applied an even-odd mask (`StableTerminalContainerView.applyGutterMask`) so tint only fills the ring area outside the inset terminal frame.
- `updateGutterTintColor()` keeps gutter tint synchronized with `resolvedTerminalBackgroundNSColor()` (including opacity/background changes), eliminating blur-vs-terminal color mismatch artifacts.

## Outside-click close logic corrected for shortcut-open path

- In `ContentView.syncStickyTerminalOutsideClickMonitor`, monitor installation now depends on:
  - notch open
  - current tab is terminal
  - not currently hovering notch
- It is no longer gated by `terminalStickyMode`, so outside-click close works even when sticky mode is disabled and terminal is opened via shortcut without hover transitions.
- Added monitor re-sync on notch open state transitions so monitor is installed even when currentView did not change during reopen.

## Custom terminal scrollbar rendering

- SwiftTerm’s built-in scroller is retained as a state source but visually hidden.
- Added `TerminalScrollKnobView` (non-intercepting overlay) rendered in `TerminalManager` as a custom rounded knob.
- KVO observers on scroller `doubleValue`, `knobProportion`, and `isEnabled` drive:
  - knob position
  - knob height
  - visibility (hidden when non-scrollable)
- Knob layout is refreshed on container resize via `refreshScrollKnobLayout()`.

image

### DIFF
--- a/DynamicIsland/ContentView.swift
+++ b/DynamicIsland/ContentView.swift
@@ -621,6 +621,11 @@ struct ContentView: View {
                 }
                 if newState == .closed {
                     removeStickyTerminalClickMonitor()
+                } else {
+                    // Install the outside-click monitor for terminal opens that don't
+                    // change `currentView` (e.g. shortcut re-opening with the terminal
+                    // tab already selected, where the cursor never enters the notch).
+                    syncStickyTerminalOutsideClickMonitor()
                 }
                 #if os(macOS)
                 if newState == .open {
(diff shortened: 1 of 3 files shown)


sharma2007:
Completely closes issue #371

Ebullioscopic (approved):

PR #512, opened 2026-06-14
Feature/fix lockscreen widgets for compatibility with new macOS 27 Siri AI overlays

S123MR:
Currently, widgets on the lock screen appear on top of the Siri AI overlay when triggering a conversation on macOS 27.
This PR implements a centralized, power-aware monitoring system for system overlays (specifically Siri) on the lock screen. It ensures that the widgets respond to Siri's activation while maintaining maximum battery efficiency .

Key Features

Centralized Siri Visibility Monitor
Introduced SiriVisibilityMonitor, a singleton that utilizes CGWindowListCopyWindowInfo to detect Siri's window state with high precision.
Provides a one-line autohide(window) helper for all window managers, centralizing the collision logic and reducing code duplication across the codebase.
Adaptive Polling Profiles (Power-Aware)
To balance "imminent" responsiveness with battery life, the monitor now dynamically shifts its heartbeat based on the Mac's power state:
High Performance (Plugged In): 30Hz active polling for near-instant detection.
Balanced (On Battery): 16Hz active polling—the "sweet spot" for fluid human perception.
Power Saver (Low Power Mode): 4Hz active polling and 0.5Hz idle polling to maximize energy longevity.
3. User-Facing Controls
Added a Siri Detection Speed picker in Settings -> Lock Screen.
Users can choose Automatic (dynamic power-based shifting) or lock the app into a specific performance profile.

## Summary by CodeRabbit

- **New Features**
  - Added a macOS 27+ “Siri Detection” option in Lock Screen settings with Automatic, High Performance, Balanced, and Power Saver modes.

- **Improvements**
  - Improved Siri and Apple Intelligence overlay detection responsiveness while reducing visibility flicker.
  - Lock Screen panels and widgets now automatically hide when Siri is visible and restore their visibility more reliably.
  - Improved visibility handling when the screen locks, wakes, or changes power state.

### DIFF
--- a/DynamicIsland/components/Settings/SettingsView.swift
+++ b/DynamicIsland/components/Settings/SettingsView.swift
@@ -5043,6 +5043,7 @@ struct LockScreenSettings: View {
     @Default(.enableLockScreenTimerWidget) private var enableLockScreenTimerWidget
     @Default(.enableLockScreenWeatherWidget) private var enableLockScreenWeatherWidget
     @Default(.enableLockScreenFocusWidget) private var enableLockScreenFocusWidget
+    @Default(.siriResponsivenessMode) private var siriResponsivenessMode
     @Default(.lockScreenWeatherWidgetStyle) private var lockScreenWeatherWidgetStyle
     @Default(.lockScreenWeatherProviderSource) private var lockScreenWeatherProviderSource
     @Default(.lockScreenWeatherTemperatureUnit) private var lockScreenWeatherTemperatureUnit
@@ -5165,6 +5166,24 @@ struct LockScreenSettings: View {
                 Text("Live Activity & Feedback")
             } footer: {
                 Text("Controls whether Dynamic Island mirrors lock/unlock events with its own live activity and audible chimes.")
+            }
+
+            Section {
+                Picker("Siri detection speed", selection: $siriResponsivenessMode) {
+                    ForEach(SiriResponsivenessMode.allCases) { mode in
+                        VStack(alignment: .leading) {
+                            Text(mode.displayName)
+                            Text(mode.description)
+                                .font(.caption)
+                                .foregroundStyle(.secondary)
+                        }.tag(mode)
+                    }
+                }
+                .settingsHighlight(id: highlightID("Siri detection speed"))
+            } header: {
+                Text("Siri Detection")
+            } footer: {
+                Text("Higher speeds allow widgets to hide almost instantly when Siri is invoked, but may impact battery life when on battery power.")
             }
 
             Section {
--- a/DynamicIsland/managers/Extensions/ExtensionLockScreenWidgetPresentationController.swift
+++ b/DynamicIsland/managers/Extensions/ExtensionLockScreenWidgetPresentationController.swift
@@ -195,7 +195,7 @@ private final class ExtensionLockScreenWidgetWindowPool {
         window.isOpaque = false
         window.backgroundColor = .clear
         window.hasShadow = false
-        window.level = .screenSaver
+        window.level = NSWindow.Level(rawValue: Int(CGWindowLevelForKey(.statusWindow)) + 1)
         window.collectionBehavior = [.canJoinAllSpaces, .stationary, .fullScreenAuxiliary]
         window.ignoresMouseEvents = true
         window.isMovable = false
--- a/DynamicIsland/managers/Extensions/ExtensionLockScreenWidgetPresentationController.swift
+++ b/DynamicIsland/managers/Extensions/ExtensionLockScreenWidgetPresentationController.swift
@@ -195,7 +195,7 @@ private final class ExtensionLockScreenWidgetWindowPool {
         window.isOpaque = false
         window.backgroundColor = .clear
         window.hasShadow = false
-        window.level = NSWindow.Level(rawValue: Int(CGWindowLevelForKey(.statusWindow)) + 1)
+        window.level = NSWindow.Level(rawValue: Int(CGShieldingWindowLevel()))
         window.collectionBehavior = [.canJoinAllSpaces, .stationary, .fullScreenAuxiliary]
         window.ignoresMouseEvents = true
         window.isMovable = false
--- a/DynamicIsland/managers/Extensions/ExtensionLockScreenWidgetPresentationController.swift
+++ b/DynamicIsland/managers/Extensions/ExtensionLockScreenWidgetPresentationController.swift
@@ -195,7 +195,7 @@ private final class ExtensionLockScreenWidgetWindowPool {
         window.isOpaque = false
         window.backgroundColor = .clear
         window.hasShadow = false
-        window.level = NSWindow.Level(rawValue: Int(CGShieldingWindowLevel()))
+        window.level = .screenSaver
         window.collectionBehavior = [.canJoinAllSpaces, .stationary, .fullScreenAuxiliary]
         window.ignoresMouseEvents = true
         window.isMovable = false
--- a/DynamicIsland/managers/LockScreenLiveActivityWindowManager.swift
+++ b/DynamicIsland/managers/LockScreenLiveActivityWindowManager.swift
@@ -21,6 +21,7 @@ import Defaults
 import SkyLightWindow
 import SwiftUI
 import QuartzCore
+import Combine
 
 @MainActor
 class LockScreenLiveActivityWindowManager {
@@ -36,6 +37,7 @@ class LockScreenLiveActivityWindowManager {
     private var screenChangeObserver: NSObjectProtocol?
     private var workspaceObservers: [NSObjectProtocol] = []
     private var currentNotchSize: CGSize?
+    private var siriVisibilityCancellable: AnyCancellable?
 
     /// Whether the target screen uses Dynamic Island (pill) mode.
     private var isDynamicIslandMode: Bool {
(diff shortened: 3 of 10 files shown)


Ebullioscopic:
Hi @S123MR can you share a screen recording of this in action?, we're all currently on Tahoe and not gonna move to macOS 27 until the stable version is up, so it'll be helpful for us to get an idea of it

S123MR:
Of course @Ebullioscopic! 
The screen recording is a bit too big so I uploaded it to Google Drive: https://docs.google.com/presentation/d/e/2PACX-1vSloKTX5cM2O8u7KyGu65KQghWYpuOhMICQD-fydMDIVYz-HbDB97WPXoa04osj_9A5DRi2tAsfcJSW/pub?start=false&loop=false&delayms=3000

here what happens with the latest release of Atoll before the fix:

Screenshot 2026-06-15 at 7 00 59 PM

Ebullioscopic:
@S123MR the core principle looks great, but I'm quite worried about having a separate polling function for monitoring this

Can you research a bit to look into some way that it can be event driven rather than polling driven (coz we already have a lot of timers running in the app)?

I'll try to find some ways too meanwhile and keep you posted of any new findings

S123MR:
Sure, I'll see

S123MR:
Hmm @Ebullioscopic, I tried a few event driven triggers such as: AXObserver, NSWorkspace etc, but they all don't seem to work due to apple's lock screen limitations or stuff like that, for now I just did a little update to the polling rates to better optimize the fluidity, but I think polling is generally ok as it won't consume CPU when unlocked or when the screen is asleep, but I'd love to know if you have any other ideas or approaches we could try!

StudioKeys (dismissed):
@S123MR So i can see some emojis that you left inside the code as well as the extensive logging could be trimmed as to reduce bottlenecks later down the line.

I'm yet to run your changesets on Tahoe, so once the changes are done, will be doing that.
Pinging me on the comments once its pushed should be enough.

S123MR:
@StudioKeys Done!

StudioKeys:
> @StudioKeys Done!

I will check this on Tahoe and see if it contradicts with anyone on Tahoe as the app is compatible with 15, 26 and 27 now

Cc: @Ebullioscopic if possible run instruments for possible crashes as well

S123MR:
Great, thx!

Ebullioscopic:
On it @StudioKeys

S123MR:
@Ebullioscopic @StudioKeys Can this be merged now?

StudioKeys (dismissed):
Had a look. The shape is right — one monitor, one `autohide()` per manager, power-aware polling. A few things to sort before it goes in though.

The big one: autohide leaks windows. The sink captures `window` strongly:

```swift
.sink { isVisible, isLocked in
    window.contentView?.layer?.removeAllAnimations()
    window.animator().alphaValue = targetAlpha
}
.store(in: &cancellables)
```

That cancellable lives in the panel managers, and those are singletons, so every window ever created stays alive for the whole session. Lock and unlock a few times and you've got a stack of dead windows still around. Worse, most callers append into a persistent `cancellables` set on each `createWindow` — Reminder, Timer, Weather, Panel all do `autohide(newWindow, cancellables: &cancellables)` and never clear it, so after a couple cycles one Siri toggle fires a whole batch of stale subscriptions animating windows that are already gone. And `LockScreenLiveActivityWindowManager` does it a third way (local set, then assigns the property), so the pattern isn't even consistent. Add `[weak window]`, bail if nil, reset the set before re-subscribing, and use the same approach in all of them.

Power source wiring is fragile too. `onPowerSourceChange` is a single closure, but `BatteryActivityManager` already exposes `addObserver`/`removeObserver(byId:)`. Assigning the closure quietly stomps anything else that wants it (or gets stomped later). Use the observer API. Separately, `isPluggedIn` starts `false` and only corrects on the first power event — seed it from current state in `init` like you already do for low power mode, otherwise a plugged-in Mac runs the battery profile until something changes.

The Hz numbers don't line up. Picker descriptions say 10 / 6.6 / 1, the code comments work out to 33 / 20 / 8 active and 10 / 6.6 / 1 idle, and the PR description says 30 / 16 / 4. Three different sets, and the picker is quoting the idle rate while the description quotes active. Pick the real ones and make all three agree.

Smaller things:

- `layer == 23`, `height > 400`, owner `"CampoRemoteService"` — move these into named constants with a note that they're macOS 27 internals and will drift on updates.
- 33Hz `CGWindowListCopyWindowInfo` on the main thread in high-performance active mode. It's gated behind locked + screen-on so it won't run all the time, but a main-thread syscall 33 times a second is worth a second look.
- Trailing whitespace on `disappearanceThreshold = 2`, plus a few stray double blank lines.
- The "Old: idle 0.20s..." comments imply replaced code, but nothing was removed and there's no prior Siri logic on main. They're misleading — drop them.

The leak is the one that blocks merge for me. The rest I'd like cleaned up but they're less urgent.

StudioKeys:
Re-reviewed the latest push. The leak's sorted — sink captures `[weak window]`, the set gets reset before each re-subscribe, and it's the same shape across all the managers now. Power source moved onto `addObserver`/`removeObserver(byId:)`, and `isPluggedIn` is seeded in init so a plugged-in Mac picks the right profile from the start instead of waiting for the first event. The Hz numbers line up now across the picker, the code, and the comments. Magic values are named constants with the macOS 27 note. Detection runs off the main thread via `Task.detached`. The misleading "Old:" comments are gone too.

One thing worth adding before this merges, and it's the compatibility angle. The monitor isn't gated by OS version. The app still ships to 15 Sequoia and 26 Tahoe, and on those the Siri overlay carries a different owner and layer, so detection never matches and the feature just sits inert. That part's fine. The problem is the timer still fires `CGWindowListCopyWindowInfo` 7-10 times a second the entire time the screen is locked, on machines where it can never hit anything. Nothing breaks, it just burns battery for no reason. Wrapping the setup in `if #available(macOS 27, *)` keeps 15 and 26 from ever starting the poll.

Smaller stuff: double blank lines around `SiriResponsivenessMode` in Constants.swift, some trailing whitespace in the manager diffs, and the new `cancellables` vars in Reminder/Timer/Weather look unused — only `siriCancellables` is referenced.

StudioKeys:
> Re-reviewed the latest push. The leak's sorted — sink captures `[weak window]`, the set gets reset before each re-subscribe, and it's the same shape across all the managers now. Power source moved onto `addObserver`/`removeObserver(byId:)`, and `isPluggedIn` is seeded in init so a plugged-in Mac picks the right profile from the start instead of waiting for the first event. The Hz numbers line up now across the picker, the code, and the comments. Magic values are named constants with the macOS 27 note. Detection runs off the main thread via `Task.detached`. The misleading "Old:" comments are gone too.
> 
> One thing worth adding before this merges, and it's the compatibility angle. The monitor isn't gated by OS version. The app still ships to 15 Sequoia and 26 Tahoe, and on those the Siri overlay carries a different owner and layer, so detection never matches and the feature just sits inert. That part's fine. The problem is the timer still fires `CGWindowListCopyWindowInfo` 7-10 times a second the entire time the screen is locked, on machines where it can never hit anything. Nothing breaks, it just burns battery for no reason. Wrapping the setup in `if #available(macOS 27, *)` keeps 15 and 26 from ever starting the poll.
> 
> Smaller stuff: double blank lines around `SiriResponsivenessMode` in Constants.swift, some trailing whitespace in the manager diffs, and the new `cancellables` vars in Reminder/Timer/Weather look unused — only `siriCancellables` is referenced.

@S123MR fix this at the earliest and also once this is done, i think it should be good to merge

Cc: @Ebullioscopic

StudioKeys:
Re-checked the follow-ups and ran a full integration pass: every point I raised is handled and the diff wires into the rest of the app cleanly. Keeping this as a comment rather than an approval so a green check doesn't nudge a merge while other lockscreen work is still in flight. Cc @Ebullioscopic for the merge call.

Ebullioscopic (dismissed):

StudioKeys:
@S123MR check the polling mechanisms and also address the reviews posted by codeRabbit

Cc: @danfq

S123MR:
> @S123MR check the polling mechanisms and also address the reviews posted by codeRabbit
> 
> Cc: @danfq

Done!
@StudioKeys

S123MR:
I'm not sure on what to check on the polling, any suggestions?

StudioKeys:
> I'm not sure on what to check on the polling, any suggestions?

Like you could test and see how is your ram and cpu being used with both Atoll running and using Siri in BG as if polling is kinda high, then there are chances of bottlenecks which can causes compatibility issues

S123MR:
Yep, of course! I've checked to ensure that the polling mechanism never initiates unless the system is in the locked state to ensure that minimal resources are consumed when working, currently the new siri ai consumes around 100-200mb of ram while inactive, but I think that apple might further optimize this in newer releases

StudioKeys:
@S123MR f you wish this PR to be merged, please rebase the PR to track the dev branch instead of main as the main branches were force-pushed to remove some LFS stuff, else you could close it yourself

Please ping either myself or @Ebullioscopic if changes are done
Cheers

S123MR:
@StudioKeys Done!

StudioKeys:
@S123MR add the changelog entries for the changes done as well

PR #531, opened 2026-06-24
fix: snap notch window frames to whole pixels and refine space behavior

fatihozdil:
## Summary

- Rounds notch window frame origins and sizes to whole pixels across creation,
  positioning, and resize paths (`DynamicIslandApp`, `MusicControlWindowManager`,
  `TimerControlWindowManager`, `ContentView`), avoiding sub-pixel blur on scaled displays
- Replaces the `NotchSpaceManager` insert/remove bookkeeping with an explicit window
  `collectionBehavior` (`.canJoinAllSpaces, .fullScreenAuxiliary, .ignoresCycle, .stationary`)
  and a leaner `styleMask`, keeping the panel visible on deactivate; deletes the now-unused
  `NotchSpaceManager` and its usages
- Reworks fullscreen detection so notch hiding is driven entirely by the **Hide DynamicIsland**
  option (`.always` / `.nowPlayingOnly` / `.never`), and uses the Accessibility `AXFullScreen`
  attribute to distinguish a genuinely fullscreen window from a merely maximized/zoomed one.
  This fixes the notch failing to hide in real fullscreen on notched MacBooks (the previous
  frame-size check compared against the full screen frame, which differs from a fullscreen
  window's safe-area frame by the notch inset) and stops the notch from hiding for maximized,
  non-fullscreen media windows. Falls back to the frame-based result when Accessibility isn't granted.

## Test plan

- [x] Built via `xcodebuild` — succeeded with no new warnings
- [x] Verified branch is rebased on current `main` with no merge conflicts
- [x] Manually verified on a notched MacBook: maximized media app stays visible, true
      fullscreen hides the notch, and exiting fullscreen restores it immediately
- [x] Recorded a demo (see comments) verifying notch-over-fullscreen behavior, maximized vs. true fullscreen hiding, and recovery on exit


StudioKeys (changes requested):
Mostly clean — the pixel rounding is low-risk and does what it says. A few things before this can go in.

It's conflicting with main right now. The description says the branch is rebased with no conflicts, but GitHub disagrees, so something landed since. Needs a rebase.

The part I can't sign off on from reading is the space change. You're dropping all the `NotchSpaceManager.notchSpace.windows.insert/remove` calls and the custom `CGSSpace(level: 2147483647)` in favor of `collectionBehavior = [.canJoinAllSpaces, .fullScreenAuxiliary, .ignoresCycle, .stationary]`. The window already sits at `level = .mainMenu + 3`, so the standard collection behavior is the supported way to ride along with fullscreen spaces, and it's likely cleaner than the private-space trick. But that max-level space was what kept the notch above fullscreen apps, and the diff doesn't tell me whether it still does. Check it against a fullscreen video, a native fullscreen-space app, Mission Control, and a second display before merging.

One thing falls out of that: nothing inserts into `notchSpace` anymore, so `NotchSpaceManager` is dead — but it still spins up a private CGSSpace at max level on launch for nothing. If you're committing to the collectionBehavior approach, delete the manager and its usage instead of leaving it allocating a space nobody joins.

On the fullscreen detection change:

```swift
(app.bundleIdentifier == musicManager.bundleIdentifier || Defaults[.hideNotchOption] == .always) &&
abs(app.windowFrame.width  - app.screen.frame.width)  < 10 &&
abs(app.windowFrame.height - app.screen.frame.height) < 10
```

This quietly changes what `.always` does. It used to hide the notch for any qualifying app on the screen; now it's gated behind the window actually filling the screen. If that's intentional (only hide on real fullscreen), fine, but call it out — it's a behavior change for anyone using "always hide". Also that `< 10` is a magic number, and on a notched MacBook a fullscreen window's height can differ from `screen.frame.height` by more than 10pt depending on how the app handles the menubar inset. Test that genuine fullscreen still trips it on a notched display, otherwise the notch stops hiding.

Smaller things:

- `newX = (centerX - (window.frame.width / 2)).rounded()` rounds using the unrounded width, but the frame width gets set to `window.frame.width.rounded()`. Half a pixel off-center when the width is fractional. `resizeWindow` already does this right — rounds the width first, then derives `newX`. Match that.
- `styleMask` drops `.utilityWindow` and `.hudWindow`. Probably no visual change since it's borderless and non-opaque, but `.hudWindow` touches vibrancy, so give the notch a look.
- `collectionBehavior` ends up set twice — once in `DynamicIslandWindow.init`, then overwritten with the same set in `AppDelegate`. Pick one. And the reorder inside `DynamicIslandWindow.swift` is a no-op since it's an OptionSet.

Nothing in the diff is outright broken. It's a rebase, verify the space behavior over fullscreen, and drop the dead manager.

fatihozdil:
@StudioKeys 

https://github.com/user-attachments/assets/aeda0e98-a560-4dc6-b838-f7d4cfb07e7e


https://github.com/user-attachments/assets/56e66e5e-ce69-4f5a-9929-ce9077e028d3

StudioKeys:
Read through this — pixel rounding and the AXFullScreen check for telling maximized from real fullscreen both look right, and dropping the private CGSSpace for plain collectionBehavior is the safer call on Tahoe. One thing before merge: you've got `.canJoinAllSpaces` and `.fullScreenAuxiliary` set together, which can fight each other on whether the notch actually draws over another app's fullscreen space, and that behavior tends to differ between Sequoia 15 and Tahoe 26 — since the CGSSpace that used to force it above is gone now, worth confirming notch-over-fullscreen still holds on both before this lands.

fatihozdil:
> Read through this — pixel rounding and the AXFullScreen check for telling maximized from real fullscreen both look right, and dropping the private CGSSpace for plain collectionBehavior is the safer call on Tahoe. One thing before merge: you've got `.canJoinAllSpaces` and `.fullScreenAuxiliary` set together, which can fight each other on whether the notch actually draws over another app's fullscreen space, and that behavior tends to differ between Sequoia 15 and Tahoe 26 — since the CGSSpace that used to force it above is gone now, worth confirming notch-over-fullscreen still holds on both before this lands.

.canJoinAllSpaces covers normal spaces and .fullScreenAuxiliary covers the fullscreen-space case. They don't overlap or contend. The pairing is the standard approach for menu‑bar/notch overlays.

I have tested it on Tahoe 26 and it is working well, as you can see in the screen recording in the comments above. I don't have a chance to test it on Sequoia 15 unfortunately.

What I can do is keep the CGSSpace logic for the "Never hide" option. One thing I want to add though: the current version of the app has this as a broken feature, at least in my tests. Especially when Spotify is open, it hides the notch even though it's not fullscreen. So the hide options aren't behaving as expected, and this PR is fixing those issues.

StudioKeys:
Let's keep the CGSSpace path as a fallback for the "Never hide" option — that's the one case where the notch should stay pinned above fullscreen no matter what, so re-adding the window to `notchSpace` there makes sense. Rest looks good to me.

I'll get this tested on Sequoia 15 and Golden Gate 27 by someone before we talk merge. Once the fallback's in and those pass, we can look at merging.

Cc @Ebullioscopic — could you run Instruments on Tahoe for this? And line up someone for Sequoia + Golden Gate, since both your machine and mine are on Tahoe, which is already covered by the recordings above.

fatihozdil:
@StudioKeys 
Done. 
Added back the CGSSpace. Every other hide option keeps the CGSSpace set empty and stays on the
collectionBehavior path, so FullscreenMediaDetector can still hide it.

Ebullioscopic (approved):

PR #536, opened 2026-06-27
Fix: animated Artwork MusicKit for polling, SystemOSDManager for process spawning

Ebullioscopic:
## Summary by CodeRabbit

* **New Features**
  * Added a throttling/retry backoff for Music artwork fetching after repeated token/auth failures to improve recovery.

* **Bug Fixes**
  * Improved on-screen display suppression behavior by pausing during macOS sleep and resuming correctly on wake.
  * Hardened process detection and reduced noisy system command output.

* **Documentation**
  * Updated localization metadata to keep translated strings in sync.


StudioKeys (changes requested):
Went through both. The SystemOSDManager side is good — pausing the watcher on sleep is correct, and that `!output!.isEmpty` force-unwrap you killed was a real crasher. pgrep exit-status guards help too.

The AnimatedArtworkManager backoff doesn't actually fire though, want to flag it before merge. The reset runs on every song change:

```swift
if key == cachedKey { return cachedVideoURL }   // same song returns here

if key != cachedKey {        // always true at this point
    tokenFailureCount = 0     // wiped on every new song
}
```

The equal-key case already returned above, so the second branch is always true. `tokenFailureCount` gets zeroed on every track, and `isWithinTokenBackoff()` needs it above zero, so it never returns true. The exponential schedule never runs. The thing actually saving you is the per-song nil cache (same song won't re-fetch), but a fresh song still hits the network even with a dead token — which is the overnight scenario this PR is meant to fix.

Fix is to check backoff before the reset, and only clear the counter on success:

```swift
cachedKey = key
if isWithinTokenBackoff() { cachedVideoURL = nil; return nil }
// ... auth / search / fetch ...
// on success only:
tokenFailureCount = 0
```

OSD half is ready. The artwork half needs the reset fixed, or drop the backoff entirely and treat it as cache-only.

PR #541, opened 2026-06-28
General Improvements to Minimalistic UI

Ebullioscopic:
### Waveform, candles, disable battery

https://github.com/user-attachments/assets/b361d6a3-71d7-44e3-87a5-bbc7b6b5f6dc

image
image

## Summary by CodeRabbit

* **New Features**
  * Added configurable visualizer “candles” plus legacy/vibrant color extraction to update spectrogram/gradient styling.
  * Added a scrubbable real-time waveform experience for the music slider.
  * Added an optional minimalistic battery indicator for notch-only UI.
* **Bug Fixes**
  * Improved Dynamic Island/notch spectrogram rendering and sizing to better match the selected visualizer configuration.
  * Music visualization updates are smoother and correctly layered.
  * System HUD suppression now properly pauses on sleep and resumes after wake; artwork fetching backs off after repeated auth/token failures.
* **Chores**
  * Refined background log export behavior and reduced custom print routing.


StudioKeys (changes requested):
Went through the whole thing. The features are nice but there's a fair bit to fix before this lands, and a couple of the audio ones are real.

Two things up front that aren't code bugs but matter. The `Localizable.xcstrings` file got fully re-serialized — about 90k lines of churn for a net 71 new lines. That buries whatever strings you actually added, and it'll conflict with anything else that touches translations. Pull the file back to main's version and re-add just the new keys. Second, this PR re-touches `AnimatedArtworkManager` and `SystemOSDManager`, which are the same files as #536, so the two will collide on merge — worth folding them together. And the token backoff in `AnimatedArtworkManager` still doesn't actually do anything here, same as I noted on #536: the `if key != cachedKey` block sits right after an early return on the equal case, so it's always true and `isWithinTokenBackoff()` never gets a chance to fire.

The ones I'd block on:

The audio callback allocates now. `getSmoothedMagnitudes` builds an `NSMutableArray` of six `NSNumber`s — that's malloc plus a dozen objc_msgSends plus an ARC release — and it runs inside `audioIOProc`, which is the CoreAudio render thread. It's only hit on `callbackCount % 1000 == 0`, so it's a debug-log path rather than every callback, but you can't malloc or message ObjC on the render thread at all; that's exactly the kind of thing that causes the occasional dropout. The block around it is already pushing its luck with a buffer loop and an os_log. Take the read out of the callback, or hand back a `std::array<float,6>` through a plain C accessor.

`vDSP_biquad_CreateSetup` isn't null-checked for the new bands. It returns NULL on failure, and a NULL setup passed into `vDSP_biquad` is undefined behaviour — and again, that's on the audio thread. Guard it and skip the band if it comes back nil.

The waveform timer has two problems. `startTimer()` assigns a fresh `Timer.scheduledTimer` without invalidating the old one first, so a second `onAppear` leaves a 60fps timer running forever. And `scheduledTimer` registers in `.default` mode, which means it stops firing during any AppKit tracking loop — so the waveform can freeze mid-interaction, which is the one moment it's on screen. Invalidate first, then build the timer and add it with `RunLoop.main.add(t, forMode: .common)`.

`AnimatedArtworkManager` can hand back the wrong song's artwork. `cachedKey = key` is set before the first `await`, but `cachedVideoURL` still points at the previous track at that moment (it only gets cleared on the failure branches). A second call with the same key during that suspension hits the `key == cachedKey` fast path and returns the old URL. Set `cachedVideoURL = nil` right after `cachedKey = key` to close that window.

`exportLogs` runs `logProcess.waitUntilExit()` synchronously inside a plain `Task {}`. `log show --last 2d` can take the better part of a minute, and doing it on the cooperative pool stalls every other async task in the app for that whole time, then the zip does it again. Move it to `Task.detached(priority: .utility)` or wrap the process in a continuation with a `terminationHandler`.

Less urgent but worth doing:

`getSmoothedMagnitudes()` mutates `displayMagnitudes` as a side effect of reading, and it now has three timer callers — the new 60fps waveform plus the two 30fps spectrums. When two fire in the same frame the smoothing gets applied twice, so the waveform and the candles end up disagreeing about the same data. Split it: one timer computes and stores once per frame, and everyone else reads a snapshot that doesn't mutate.

`Logger.swift` shadows `Swift.print` across the whole app and decides error vs warning by substring, so a path or filename containing "error" gets bumped to error level. It also reads `Defaults[.logLevel]` on every print. Gate your own `Logger.log()` on the level and leave `Swift.print` alone.

The `minimumInterval: 1.0` got dropped from the `TimelineView` in `LockScreenMusicPanel` and `MinimalisticMusicPlayerView`. Those are second-resolution progress bars, so now they re-lay-out at display refresh rate for no visible gain. The removal in `NotchHomeView` is fine since the scrubber needs frame-rate updates — just restore the cap on the other two.

On the C++ side, `getBand(int)` has no bounds check and `AudioBridge.mm` hardcodes `6` while `kBands` is private, so the two are coupled by luck. Expose a band-count accessor and assert the index.

Minor stuff: `NSImage.prominentOpposingColors` reads `Defaults[.colorExtractionMode]` inside the background closure, so a mid-computation toggle could mix paths — read it once before dispatching. In `ContentView`, the `DynamicIslandHeader` if/else renders the same view in both branches and only differs by a `.frame`, so make it a conditional modifier instead. And the `#import <simd/simd.h>` in `AudioBridge.h` is dead now that the simd return type is gone.

The CI workflow change is fine, good even — the `push` trigger syncs the mirror on every push with the release steps gated correctly, and `persist-credentials: false` is a small hardening win. Just confirm the `dev` branch actually exists.

Ebullioscopic:
@coderabbitai summarize the changes needed

PR #544, opened 2026-06-30
Fix stats popovers for localized graph titles

swording-k:
## Summary

Fixes stats detail popovers so they are keyed by a stable `ProcessRankingType` instead of the localized graph title string.

## Root cause

`NotchStatsView` used the displayed graph title to decide whether a stats card was clickable and which popover binding to open:

- `"CPU"`
- `"Memory"`
- `"GPU"`
- `"Network"`
- `"Disk"`

Those titles are localized before rendering. In non-English locales, for example Simplified Chinese, `Memory` becomes `内存`, so checks such as `graphData.title == "Memory"` no longer match. That makes localized stats cards lose the hover hint and can prevent the correct detail popover from opening.

## What changed

- Added an optional `rankingType` to `GraphData`.
- Assigned stable ranking types when constructing CPU, Memory, GPU, Network, and Disk graph data.
- Updated click handling, popover bindings, hover state, and the "Click for details" hint to use `rankingType` instead of localized display text.
- Kept the localized title only for UI display.

## User impact

This should restore consistent stats card behavior in localized environments. For example, in Simplified Chinese, the `内存`, `网络`, and `磁盘` cards can still show the detail hint and open the correct detail popover.

## Validation

- Reproduced the old string-matching failure with a localized memory title (`内存`).
- Built the app successfully:

```bash
xcodebuild -project DynamicIsland.xcodeproj -scheme DynamicIsland -configuration Debug -derivedDataPath /tmp/AtollDerivedData CODE_SIGNING_ALLOWED=NO build
```

Result: `BUILD SUCCEEDED`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved stats card interactions so clicks and hover behavior now open the correct details more reliably.
  * The “Click for details” hint now appears more consistently for supported charts.
  * Updated chart navigation to work across CPU, Memory, GPU, Network, and Disk views without relying on title text.

### DIFF
--- a/DynamicIsland/components/Notch/NotchStatsView.swift
+++ b/DynamicIsland/components/Notch/NotchStatsView.swift
@@ -25,6 +25,7 @@ protocol GraphData {
     var color: Color { get }
     var icon: String { get }
     var type: GraphType { get }
+    var rankingType: ProcessRankingType? { get }
 }
 
 enum GraphType {
@@ -40,6 +41,7 @@ struct SingleGraphData: GraphData {
     let color: Color
     let icon: String
     let type: GraphType = .single
+    let rankingType: ProcessRankingType?
 }
 
 // Dual value graph data (for network/disk)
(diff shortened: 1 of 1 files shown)


Ebullioscopic:
@coderabbitai review this

Ebullioscopic (approved):

PR #553, opened 2026-07-02
fix: snap song slider on notch open so it stops jumping back and slow-animating

### LINKED ISSUE #506: Song slider jumps back on hover before correcting itself (animation too slow)
**Issue Description**
When hovering over the notch UI while a song is playing, the song progress slider does not stay aligned with the actual playback time.
Instead:
• The slider jumps backward for a moment,
• Then slowly animates forward to the correct position,
• Causing a noticeable delay before it syncs with the real playback time.
This doesn’t seem like a bug, but the animation duration is too long, making the UI feel unresponsive.

**Expected Behavior**
The slider should:
• Either update in real time with no backward jump,
or
• Use a much faster correction animation so the user doesn’t have to wait for it to catch up.

**Actual Behavior**
• Hovering triggers the slider to reset backward.
• It then animates forward slowly to the correct timestamp.
• This delay makes it hard to check or scrub the song position accurately.

**Steps to Reproduce**
1. Play any song in Atoll.
2. Hover over the notch UI while the track is playing.
3. Observe the slider jump backward and then slowly animate to the correct time.

**Environment**
• macOS 15 Sequoia
• Atoll (latest version)

mvanhorn:
## Summary

Fixes the song slider jumping back and then slowly animating forward each time the notch opens (issue #506). `MusicSliderView` now snaps to the current playback position on appear without animation, and periodic timeline updates no longer inherit an animation, so the handle sits at the right spot immediately instead of catching up.

## Why this matters

When the notch reopens, the slider was mounted at a stale value and then animated toward the real playback position, which read as the handle "jumping back" and crawling forward. The reporter in #506 describes exactly this: the slider resets and slow-animates on every open. The root cause is that the view did not sync its value to the current position on appear, and the periodic progress updates ran through the default implicit animation.

This snaps to the current position on appear (skipping the snap for live streams and while the user is actively dragging, so those paths are unaffected) and disables the inherited animation for timeline progress updates. Behavior while dragging and for live streams is unchanged.

## Testing

macOS SwiftUI change verified by inspection against the existing `MusicSliderView` update path; the project has no test target to run. Confirmed the snap is guarded by the live-stream and active-drag checks so it only affects the reopen case described in the issue.

Fixes #506

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the music slider’s behavior so its position updates smoothly without unwanted animations.
  * Set the slider’s starting position more reliably when the view appears, while avoiding changes during active dragging or live streams.

### DIFF
--- a/DynamicIsland/components/Notch/NotchHomeView.swift
+++ b/DynamicIsland/components/Notch/NotchHomeView.swift
@@ -801,10 +801,15 @@ struct MusicSliderView: View {
                 }
             }
         }
+        .onAppear {
+            guard !isLiveStream else { return }
+            guard !dragging else { return }
+            setSliderValueWithoutAnimation(MusicManager.shared.estimatedPlaybackPosition())
+        }
         .onChange(of: currentDate) { newDate in
             guard !isLiveStream else { return }
             guard !dragging, timestampDate.timeIntervalSince(lastDragged) > -1 else { return }
-            sliderValue = MusicManager.shared.estimatedPlaybackPosition(at: newDate)
+            setSliderValueWithoutAnimation(MusicManager.shared.estimatedPlaybackPosition(at: newDate))
         }
         .onChange(of: isPlaying) { _, playing in
             // Snap slider to the exact position when music pauses so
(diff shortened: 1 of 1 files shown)


Ebullioscopic (approved):

mvanhorn:
Cheers and thanks, @Ebullioscopic - the song slider staying put on notch open was a satisfying fix.

Issue #463, opened 2026-05-18
[Bug] Calendar widget fails to autoscroll to current event when all-day reminders are present


dralan71:
**Current Behavior**
When an all-day reminder or event is present in the calendar widget, the view does not autoscroll to the current time-based event. The widget ends up pinned to the top of the day, showing only the all-day item and the first scheduled event. This requires manual scrolling to see what is currently happening or coming up next.

**Expected Behavior**
- All-day events and reminders should remain visible at the top of the widget.
- Time-constrained (scheduled) events should continue to autoscroll to the current or next upcoming event based on the time of day.
- The user should not need to manually scroll to find their current event just because an all-day item exists.

**Suggested Solution / Implementation Idea**
Split the calendar widget into two separate areas:
1. A pinned/static section at the top for all-day events and reminders.
2. An independent scrollable section below for time-scheduled events, which handles autoscrolling to the current event.

This would decouple the all-day items from the timed scroll logic, ensuring autoscroll continues to work for the events that actually have a start time.

**Steps to Reproduce**
1. Add an all-day reminder or event to the calendar.
2. Add one or more time-specific events throughout the same day.
3. Open the calendar widget at a time when a later event is current.
4. Observe that the widget does not scroll to the active time-specific event, and instead stays at the top near the all-day item.

Using version 2.1.0 (11) with Auto-scroll to next event enabled

PR #566, opened 2026-07-07
Fix calendar auto-scroll when all-day events/reminders are present

swording-k:
# PR: Fix calendar auto-scroll when all-day events/reminders are present

**Closes:** #463

## Problem

When all-day events or reminders are present in the calendar widget, the auto-scroll feature fails to position the view at the current/next timed event. Instead, the view stays pinned to the top of the day, showing only all-day items and early morning events. Users must manually scroll down to see what's currently happening.

## Root Cause

Three issues in `EventListView.scrollToRelevantEvent()`:

1. **All-day events block auto-scroll**: The scroll target selection (`nonAllDayUpcoming ?? firstAllDay ?? lastEvent`) falls back to `firstAllDay` when no timed event matches, scrolling to the top of the list where all-day items sit — making the view appear stuck at the top.

2. **Reminders distort in-progress detection**: Non-all-day reminders have `end = endOfDay` (23:59:59), so the condition `$0.end > now` incorrectly matches reminders as "in-progress" for the entire day, preventing scroll to the actual current timed event.

3. **No separation between all-day and timed events**: All events (all-day + timed) are mixed in a single scrollable list. All-day items always appear first (start = 00:00), and scrolling past them requires manual effort.

### 1. Split event list into pinned all-day section + scrollable timed section

- **All-day events/reminders** are displayed in a pinned section at the top (does not scroll). A thin separator line visually distinguishes the sections.
- **Timed events** are in an independent scrollable list below that handles auto-scrolling.
- This decouples all-day items from the timed scroll logic, ensuring auto-scroll always targets timed events.

### 2. Improve scroll target selection for timed events

New priority order (only from timed events):
1. **In-progress timed event** (`start <= now && end > now`)
2. **Next upcoming timed event** (`start > now`)
3. **Last timed event** (fallback)

No fallback to all-day events — they're already visible at the top.

### 3. Fix reminder in-progress detection

Reminders are point-in-time events, not duration events. Their `end = endOfDay` is an artificial value that shouldn't be used for in-progress checks. The fix treats non-all-day reminders as "in-progress" only within 1 hour of their `start` time, rather than for the entire day based on `end`.

### 4. Auto-scroll triggers only on appear and date change, not on every data refresh

Previously, `.onChange(of: timedEvents)` would force the scroll position back to the current event every time calendar data refreshed — this locked the view and prevented manual scrolling. Now, auto-scroll triggers only on `.onAppear` (first open) and `.onChange(of: selectedDate)` (date switch via WheelPicker). After the initial positioning, users can freely scroll up/down to view all events.

### 5. Add auto-scroll to StandaloneEventCardList

The standalone calendar view (`StandaloneCalendarView`) was missing auto-scroll entirely. The same pinned-all-day + scrollable-timed pattern is applied, with `ScrollViewReader` and `scrollToRelevantEvent` logic matching `EventListView`.

## Test Plan

1. Create an all-day event and several timed events spread across the day
2. Open the calendar widget at a time when a later timed event is current
3. Verify: all-day event visible at the top, timed list auto-scrolls to current event
4. Create a timed reminder (e.g. 9:00 AM) and verify it doesn't prevent scroll to the actual current event
5. Create only all-day events with no timed events — verify all-day items display normally at top
6. Toggle "Auto-scroll to next event" off — verify no auto-scroll occurs
7. Toggle "Hide all-day events" on — verify all-day section disappears, timed list scrolls normally
8. Build succeeds (`xcodebuild` with no signing — verified ✅)

## Files Changed

- `DynamicIsland/components/Calendar/DynamicIslandCalendar.swift` — EventListView, StandaloneEventCardList

## Summary for reviewers (follow-up fixes on this branch)

Please review and merge the **final code at the branch tip (`fa45a32`)**. The cumulative diff from the base is what matters — earlier commits on this branch were intermediate iterations that each had a problem, now all resolved:

- **Original approach (base commits):** all-day events were pinned as a vertical stack — one full-height row per event. With two all-day events, the panel was completely filled, hiding the timed-event scroll area (the #463 follow-up regression).
- **First follow-up (`d5d46e3`):** fixed the compact strip for the Dynamic Island panel only; the standalone calendar window still rendered all-day events as full-height cards (tree-like), so two all-day events still filled that window.
- **Second follow-up (`95da501`):** added a one-shot auto-scroll retrigger so the initial scroll still fires when `timedEvents` load after the view appears (addressed a CodeRabbit review finding).
- **Final commit (`fa45a32`):** applied the same compact single-row strip to the standalone calendar window too, making both views consistent.

**Net result (merge-ready state):** both the Dynamic Island panel and the standalone calendar now render all-day events as a compact single-row strip (like macOS Calendar's Day view). The timed-event scroll area always retains its full height regardless of how many all-day events exist, and auto-scroll works even when timed events load late. CI is green on both macOS 15 and macOS 26.

A squash-merge is fine — the branch tip reflects the correct final state.

## Summary by CodeRabbit

* **New Features**
  * Calendar views now pin all-day events in a compact single-row strip, with timed events shown separately in a scrollable list.
  * Enhanced date switching jumps to a more relevant event, with improved reminder-aware navigation.
  * All-day event chips can include reminder completion toggles and open the system calendar when tapped.

* **Bug Fixes**
  * Improved, more reliable auto-scrolling behavior when changing the selected date, with better synchronization across the event list and standalone cards.

### DIFF
--- a/DynamicIsland/components/Notch/NotchHomeView.swift
+++ b/DynamicIsland/components/Notch/NotchHomeView.swift
@@ -723,9 +723,6 @@ struct NotchHomeView: View {
                 if shouldShowMusicPlayer {
                     MusicPlayerView(albumArtNamespace: albumArtNamespace)
                         .frame(maxWidth: .infinity, alignment: .leading)
-                        // Cap height so the album art (aspectRatio 1) doesn't grow
-                        // with the calendar column — keeps controls inside the notch.
-                        .frame(maxHeight: 158)
                 }
                 
                 if Defaults[.showCalendar] {
--- a/DynamicIsland/components/Notch/NotchHomeView.swift
+++ b/DynamicIsland/components/Notch/NotchHomeView.swift
@@ -723,6 +723,9 @@ struct NotchHomeView: View {
                 if shouldShowMusicPlayer {
                     MusicPlayerView(albumArtNamespace: albumArtNamespace)
                         .frame(maxWidth: .infinity, alignment: .leading)
+                        // Cap height so the album art (aspectRatio 1) doesn't grow
+                        // with the calendar column — keeps controls inside the notch.
+                        .frame(maxHeight: 158)
                 }
                 
                 if Defaults[.showCalendar] {
--- a/DynamicIsland/models/Constants.swift
+++ b/DynamicIsland/models/Constants.swift
@@ -1301,21 +1301,13 @@ extension Defaults.Keys {
     static let enableAppleNotesSync = Key<Bool>("enableAppleNotesSync", default: false)
     static let appleNotesLastSyncDate = Key<Date?>("appleNotesLastSyncDate", default: nil)
     
-    // Default media controller.
-    //
-    // Use the system "Now Playing" (MediaRemote) controller by default: it
-    // reads the macOS Now Playing target and therefore works with EVERY media
-    // app — Apple Music, NetEase Cloud Music (网易云音乐), Spotify, QQ Music,
-    // etc. Apple Music also registers with the system Now Playing, so picking
-    // Now Playing does NOT lose Apple Music support; it merely restores the
-    // third-party players that the Apple-Music-only controller cannot see.
-    //
-    // Previously the default was switched to `.appleMusic` on macOS 15.4+, which
-    // silently broke recognition of NetEase Cloud Music and other apps (see
-    // Settings footer: "'Now Playing' ... works with all media apps"). Restored
-    // to `.nowPlaying` so all players are recognized out of the box.
+    // Helper to determine the default media controller based on macOS version
     static var defaultMediaController: MediaControllerType {
-        .nowPlaying
+        if #available(macOS 15.4, *) {
+            return .appleMusic
+        } else {
+            return .nowPlaying
+        }
     }
     
     // Migration helper to convert from legacy enableGradient Boolean to new ProgressBarStyle enum
--- a/DynamicIsland/models/Constants.swift
+++ b/DynamicIsland/models/Constants.swift
@@ -1301,13 +1301,21 @@ extension Defaults.Keys {
     static let enableAppleNotesSync = Key<Bool>("enableAppleNotesSync", default: false)
     static let appleNotesLastSyncDate = Key<Date?>("appleNotesLastSyncDate", default: nil)
     
-    // Helper to determine the default media controller based on macOS version
+    // Default media controller.
+    //
+    // Use the system "Now Playing" (MediaRemote) controller by default: it
+    // reads the macOS Now Playing target and therefore works with EVERY media
+    // app — Apple Music, NetEase Cloud Music (网易云音乐), Spotify, QQ Music,
+    // etc. Apple Music also registers with the system Now Playing, so picking
+    // Now Playing does NOT lose Apple Music support; it merely restores the
+    // third-party players that the Apple-Music-only controller cannot see.
+    //
+    // Previously the default was switched to `.appleMusic` on macOS 15.4+, which
+    // silently broke recognition of NetEase Cloud Music and other apps (see
+    // Settings footer: "'Now Playing' ... works with all media apps"). Restored
+    // to `.nowPlaying` so all players are recognized out of the box.
     static var defaultMediaController: MediaControllerType {
-        if #available(macOS 15.4, *) {
-            return .appleMusic
-        } else {
-            return .nowPlaying
-        }
+        .nowPlaying
     }
     
     // Migration helper to convert from legacy enableGradient Boolean to new ProgressBarStyle enum
--- a/DynamicIsland/components/Calendar/DynamicIslandCalendar.swift
+++ b/DynamicIsland/components/Calendar/DynamicIslandCalendar.swift
@@ -716,18 +716,15 @@ private struct StandaloneEventCardList: View {
 
     var body: some View {
         VStack(spacing: 0) {
-            // Pinned all-day section
+            // Compact all-day strip — a single fixed-height row of small chips
+            // (like macOS Calendar's Day view), so multiple all-day events no
+            // longer stack into full rows and consume the whole panel (#566).
             if !allDayEvents.isEmpty {
-                VStack(spacing: 8) {
-                    ForEach(allDayEvents) { event in
-                        eventCard(event)
-                    }
-                    Rectangle()
-                        .fill(Color.gray.opacity(0.2))
-                        .frame(height: 1)
-                        .padding(.horizontal, 2)
-                }
-                .padding(.vertical, 2)
+                AllDayEventsStrip(
+                    events: allDayEvents,
+                    onToggleReminder: onToggleReminder
+                )
+                .padding(.vertical, 4)
             }
 
             // Scrollable timed events with auto-scroll
--- a/DynamicIsland/components/Calendar/DynamicIslandCalendar.swift
+++ b/DynamicIsland/components/Calendar/DynamicIslandCalendar.swift
@@ -60,6 +60,25 @@ func scrollReferenceTime(for date: Date) -> Date {
 
 // MARK: - Compact all-day events strip
 
+/// Shared layout metrics for the all-day events strip.
+///
+/// The strip height and the space reserved for it in the timed-events list must
+/// stay in sync across `AllDayEventsStrip`, `StandaloneEventCardList`, and
+/// `EventListView`. Centralising the values here avoids the hardcoded
+/// `28` / `3` / `41` / `30` duplicated in multiple views (#566 review).
+private enum AllDayStripMetrics {
+    /// Height of the horizontal chip row inside `AllDayEventsStrip`.
+    static let chipRowHeight: CGFloat = 28
+    /// Vertical padding applied around the chips and the strip content.
+    static let verticalPadding: CGFloat = 3
+    /// Top inset pushed onto the timed `List` so `scrollTo(.top)` lands the
+    /// target event fully below the floating all-day overlay
+    /// (chip row + 1pt divider + 1pt margin ≈ 30pt). Shared by `EventListView`
+    /// and `StandaloneEventCardList` so both reserve exactly the same space.
+    /// (#566 review: de-duplicate the hardcoded 28/3/41/30)
+    static let listTopInset: CGFloat = 30
+}
+
 /// Horizontal, single-row strip of all-day events.
 ///
 /// The Dynamic Island calendar panel is height-constrained (`CalendarView` is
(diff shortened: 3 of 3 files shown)


swording-k:
Follow-up fix for the all-day events layout (addresses a regression introduced while fixing #463):

The previous version pinned all-day events as a vertical stack in the top section. In the height-constrained Dynamic Island calendar panel (capped at a fixed height), two all-day events could fill the entire panel, leaving no room for the timed-events scroll area — so other time-based events were hidden and the view could not scroll.

This commit replaces that growing pinned section with a single-row horizontal chip strip (`AllDayEventsStrip`). It renders all all-day events as compact chips and scrolls horizontally when overflowing, keeping the section at a constant height regardless of how many all-day events exist. The timed-events scroll area below now always retains its full height, so auto-scroll and other events stay visible even with multiple all-day events.

- Reminder completion toggles are preserved inside the chips.
- Only the Dynamic Island panel (`EventListView`) is changed; the larger standalone calendar view already has ample height and keeps its full-row all-day rendering.

swording-k:
Addressed the CodeRabbit review finding (auto-scroll needs an events-based retrigger):

- The initial auto-scroll previously only fired from `.onAppear` and `onChange(of: selectedDate)`. If `timedEvents` became available asynchronously after the view appeared, the first scroll could be missed and the view stayed at the top.
- Added a one-shot trigger: `.onChange(of: timedEvents.isEmpty)` fires the auto-scroll once when `timedEvents` transitions from empty to non-empty, guarded by a `@State private var initialAutoScrollDone` flag so it does not re-scroll on every later data refresh.
- Applied to both `EventListView` (Dynamic Island panel) and `StandaloneEventCardList` (standalone calendar), since both used the same weak trigger.

Commit: 95da501. Only `DynamicIslandCalendar.swift` changed.

swording-k:
Additional fix pushed (commit fa45a32): the standalone calendar window (`StandaloneEventCardList`) still rendered all-day events as one full-height card per event, so two all-day events filled the whole panel. Reused the same `AllDayEventsStrip` (single fixed-height row of small chips, like macOS Calendar's Day view) used by the Dynamic Island panel, so the layout is now consistent across both views and multiple all-day events never consume the page.

This addresses the `还是树状排列` / `两个全天事件占满页面` regression. Combined with the earlier commits, both the Dynamic Island panel and the standalone window now show all-day events as a compact strip.

Branch tip: fa45a32. Only `DynamicIslandCalendar.swift` changed.

Ebullioscopic:
@swording-k I got some issues with the implementation
image
Firstly, the concept of having different vertical scroll and horizontal scroll for differentiating between `All Day Events` is the good idea, but it has come up with some issues per se:

- The normal events are perpetually inside a very small hit box to scroll
- The shadow overlay is on them all the time
- Finally, when All day events are there, other events are not visible, especially if it is a meeting where the `Join Meeting` button is there

Screenshot 2026-07-09 at 8 08 04 PM

I hope these issues can be resolved, ping me once you're done, I'll re-review it and merge it if everything's good

swording-k:
Hi @Ebullioscopic,

Thanks for the detailed feedback! I've pushed fixes for all three issues you raised:

1. **Timed-events scroll area too small** — The all-day strip no longer eats vertical space. It's now rendered as a floating overlay on top of the full-height timed-events scroll area (with a reserved spacer so events aren't hidden behind it), so the scroll region keeps the entire panel height.

2. **Shadow overlay always visible** — Removed the always-on `LinearGradient` overlays from both the Dynamic Island panel and the standalone calendar window. No more permanent shadow on top of the events.

3. **All-day events hiding the Join Meeting button** — Auto-scroll now prioritizes an in-progress event that carries a conference (Join) link, so its Join button is scrolled into view first. The original "scroll to the event nearest the current time" behavior is preserved as the fallback.

Also included some polish: bumped the calendar panel height a bit and removed the unused gradient padding so more timed events are visible at once, and capped the music + calendar row to the expanded notch height so the media player controls stay fully inside the Dynamic Island bounds (previously the taller calendar could push them out).

Please take another look when you get a chance — happy to tweak anything. 🚀

DynamicIsland/models/Constants.swift
```diff
@@ -1301,13 +1301,21 @@ extension Defaults.Keys {
...
+    // silently broke recognition of NetEase Cloud Music and other apps (see
+    // Settings footer: "'Now Playing' ... works with all media apps"). Restored
+    // to `.nowPlaying` so all players are recognized out of the box.
     static var defaultMediaController: MediaControllerType {
-        if #available(macOS 15.4, *) {
-            return .appleMusic
-        } else {
-            return .nowPlaying
-        }
+        .nowPlaying
```
StudioKeys: This deviates from the original intention of PR that you have put up?

DynamicIsland/components/Notch/NotchHomeView.swift
```diff
@@ -723,6 +723,9 @@ struct NotchHomeView: View {
                 if shouldShowMusicPlayer {
                     MusicPlayerView(albumArtNamespace: albumArtNamespace)
                         .frame(maxWidth: .infinity, alignment: .leading)
+                        // Cap height so the album art (aspectRatio 1) doesn't grow
```
StudioKeys: Comments could be trimmed out

StudioKeys (changes requested):
I will post a detailed review once i go through it and test it on my machine as for now, these could be looked upon

Ebullioscopic:
Hi @swording-k I reviewed the PR again, and this time, the implementation looks cleaner but has some consistency issues:

image

- Firstly, the height of the notch view has increased, this has caused a lot of inconsistent spacing when we look at the media player
- Secondly, when you hover upon the notch, you can see that it detaches from the top for a brief moment (a lighter wallpaper might help you out to see it more easily)
- The shadows in the scroll for the date switching scroll is missing

I hope these issues can be resolved. You can ping me or @StudioKeys once the PR is ready for review

swording-k:
## Summary of this round

Pushed commit `17cc914` to address review #2 feedback plus a few refinements. Highlights:

**Review #2 fixes (consistency)**
- **Media-player spacing**: reverted the accidental `MusicPlayerView` `.frame(maxHeight: 158)` and the calendar bump (158 → 120). The expanded notch window stays its original fixed size; we now reuse internal space instead of growing children, so the music player proportions are back to normal.
- **Hover detach**: because the content height no longer changes during the expand animation, the window anchor stays put — no gap at the menu bar on hover.
- **Date-scroll shadows**: restored the `WheelPicker` left/right edge-fade gradients (they were deleted by mistake earlier). They now appear only while the strip is scrollable (on hover), which also matches the "non-permanent" intent from review #1.

**UX refinements**
- **Collapsible date header**: by default the header shows just a compact `weekday, day` chip; hovering expands it to reveal the month/year + horizontal date strip. This frees vertical room for the event list.
- **Floating all-day overlay**: all-day events now render as a floating overlay on top of the timed-event list (zero vertical cost) instead of consuming a flow row, so 2–3 timed events stay visible even when an all-day event exists.
- **Mouse drag on the date strip**: added press-and-drag scrubbing on the `WheelPicker` (in addition to trackpad two-finger scroll). No arrow buttons, per your preference.
- **Correct nearest-event auto-scroll**: the event list is now sorted chronologically and auto-scrolls to the in-progress / next-upcoming event; `scrollTo` uses a top anchor with list top-padding so the targeted event is never clipped behind the all-day overlay.

Built and verified locally (Build 14). Ready for another look when you have a moment.

StudioKeys (dismissed):
1. **StandaloneEventCardList still uses LazyVStack in ScrollView instead of List** - This is inconsistent with EventListView which uses List. The List version handles scroll position restoration better and has built-in separators. Consider unifying them both.

2. **EventModel properties used but not in the files I read**: The code uses event.conferenceURL and event.calendarAppURL() - these are from CalendarModel.swift extensions. The PR assumes these exist which is fine, but worth verifying they're stable.

3. **Magic numbers in AllDayEventsStrip**:
.frame(height: 28)  // chip height
.padding(.vertical, 3)
// plus 1pt divider + 4pt padding = ~41pt total
These should be constants, especially since the same 41 is hardcoded in EventListView (.padding(.top, 30)) and StandaloneEventCardList (.frame(height: 41)).

4. **StandaloneEventCardList missing hideAllDayEvents / hideCompletedReminders filtering** - It receives pre-filtered events but the filtering logic lives in StandaloneCalendarView. This is fine architecturally but worth noting the filteredEvents computed property in StandaloneCalendarView doesn't pass through hideAllDayEvents / hideCompletedReminders to the strip (though the strip reads from allDayEvents which is already filtered).

5. **scrollReferenceTime uses Calendar.current.isDateInToday** - This creates a hard dependency on system calendar. Consider injecting Calendar for testability (though in SwiftUI/AppKit this is less critical).

6. **Typo in Constants.swift diff**: var LocalizedName : String { - capital L, space before colon. This is a style violation.

7. **The Constants.swift changes are unrelated to the PR** - They remove cider from media controller and change default media controller to .nowPlaying. These should be in a separate PR to keep changes focused. Would be fixed if u merge latest main branch from upstream

I hope these could be addressed soon, and then we will have a check on this PR again.
All these are meant to have maintainable code instead of hardcoding and creating a mess later.

swording-k:
Thanks for the detailed review, @StudioKeys — everything actionable is addressed in the latest push (`095d11f`):

1. **List unification** — `StandaloneEventCardList` now uses `List` (was `LazyVStack` inside `ScrollView`), matching `EventListView`. It shares the same floating all-day overlay styling and the same `listTopInset` reservation, so the two views stay consistent.
2. **`conferenceURL` / `calendarAppURL()`** — confirmed they come from the `CalendarModel` extensions and are stable; no change needed.
3. **Magic numbers** — extracted the hardcoded `28 / 3 / 41 / 30` into a single `AllDayStripMetrics.listTopInset` constant shared by `AllDayEventsStrip`, `StandaloneEventCardList`, and `EventListView`, so the strip height and the reserved space can never drift apart again.
4. **`hideAllDayEvents` / `hideCompletedReminders` filtering** — left as-is; the strip reads from `allDayEvents`, which is already pre-filtered, as you noted.
5. **`scrollReferenceTime` using `Calendar.current`** — left as-is for now (you flagged it as less critical); can inject `Calendar` later if you'd prefer.
6. **`var LocalizedName` typo** — that one actually pre-exists on upstream `main` (it's in the base blob, not introduced by this PR), so I deliberately left it untouched to keep this PR focused on #463. Happy to fix it in a separate PR if you'd like.
7. **`Constants.swift` out-of-scope change** — reverted entirely to the upstream base; the media-controller default now matches `main` again.

Also fixed a latent bug while unifying: `StandaloneEventCardList.timedEvents` wasn't sorted by start, so "scroll to nearest event" could pick the wrong target — it's now sorted like `EventListView`.

Ready for another pass whenever you have time. 🙏

StudioKeys:
Looks good from my end, merges could be done by @Ebullioscopic upon a final pass of reviews

Ebullioscopic:
> Looks good from my end, merges could be done by @Ebullioscopic upon a final pass of reviews

Looks good from my end too, approved

Ebullioscopic (approved):

PR #579, opened 2026-07-10
Add Spotify 'Like Song' media control (save to Liked Songs from the notch)

fgoos:
## What

Adds an optional **Like Song** control to the configurable media control slots: a heart button that shows whether the current Spotify track is in the user's Liked Songs and toggles it with one click — no more switching to the Spotify app to save a song. Works in the standard notch player, minimalistic mode, and the lock screen panel.

## Why

Saving the currently playing track is a very common action while listening to playlists/radio, and it's the one playback action that currently still requires leaving the notch.

## Implementation notes

- **Auth:** Uses the official Spotify Web API via OAuth 2.0 PKCE (`user-library-read` + `user-library-modify`), implemented in a new `SpotifyLibraryManager`. The existing `sp_dc` Canvas session could not be reused: Spotify now rejects web-player tokens on `api.spotify.com` across the board (constant HTTP 429 with `retry-after: 34`, even with a client-token header — this likely also affects the existing `/v1/tracks` fallback in `SpotifyController`).
- **API:** Uses the new February 2026 library endpoints (`PUT/DELETE /v1/me/library?uris=…` and `GET /v1/me/library/contains?uris=…`). The legacy `/v1/me/tracks` save/contains endpoints silently return 403 for newly created apps.
- **Setup:** Users connect their own (free) Spotify developer app via a new settings section — an open-source app can't ship a shared client secret, and PKCE keeps the flow secret-free. Redirect URI is captured via `ASWebAuthenticationSession`, no Info.plist scheme registration needed.
- **UI:** New `.likeTrack` case follows the existing source-exclusive pattern (`isSpotifyExclusive`, analogous to `airPlay`'s `isAppleMusicExclusive`) — the slot renders empty for non-Spotify sources and the picker disables it. State lives in `MusicManager` (`isCurrentTrackLiked: Bool?`, nil = unknown → button disabled), fetched once per track change, with optimistic toggle and rollback on API failure.

## Testing

Built and in daily use on macOS 26.5. Verified: heart reflects saved state per track; like/unlike round-trips visible in the Spotify client; control hidden for Apple Music/YouTube Music; disabled (not broken) when no account is connected; existing controls unaffected.

## Summary by CodeRabbit

* **New Features**
  * Added a Spotify “Like Song” (heart) control to music panels.
  * Added Spotify library connection settings (Client ID, connect/disconnect, auth status).
  * Like status is shown in the UI with optimistic like/unlike; the button is disabled when the state is unavailable.
  * Spotify-only controls now appear only when Spotify is active.
* **Bug Fixes**
  * Improved Apple Music/Spotify activity detection to consistently drive slot layout and AirPlay refresh behavior.
* **Tests**
  * Added a dedicated test target for Spotify library functionality.


fgoos:
@coderabbitai Thanks for the review — all four points addressed in f4ef737:

1. **SecRandomCopyBytes guard:** result is now checked; on failure the connect flow aborts with a user-facing error instead of using predictable PKCE material (chose a graceful abort over `fatalError`).
2. **Coalesced token refresh:** concurrent `validAccessToken` calls now share a single in-flight refresh task, so rotated refresh tokens can no longer race.
3. **Keychain storage:** the OAuth token pair moved from Defaults into the Keychain (`kSecClassGenericPassword`, service-scoped), with a one-time migration for early builds; client ID and expiration remain in Defaults as suggested.
4. **Deduplicated `isAppleMusicActive`/`isSpotifyActive`:** hoisted onto `MusicManager` and referenced from all four slot views.

Note: a few further pre-existing `isAppleMusicActive` duplicates exist in unrelated components (e.g. SettingsView) that predate this PR — left untouched to keep the diff scoped, happy to consolidate in a follow-up if desired.

StudioKeys:
@fgoos attaching some sort of screenshots or screen recordings so that we could proceed to verify before even looking at the code, just for reference

fgoos:
@StudioKeys thanks for taking a look! I couldn't grab a screen recording right this second, so here are the key states as screenshots — happy to record a full **like → unlike round-trip** (heart filling/emptying + the track appearing in Liked Songs in the Spotify client) if that's still useful for verification.

**1. Setup / connection** — the *Spotify Like Button* settings section after connecting an app, showing the OAuth flow reaching "Connected — like button ready". (The Client ID is blacked out — it's public by design in the PKCE flow, but no reason to broadcast it.)

**2. Live control** — the heart in the notch player; green/filled = the current track is in Liked Songs.

**3. Configuration** — the "Like Song" control in the palette, draggable into any media slot (shown here placed into the layout preview).

One thing to preempt: the heart renders **Spotify-green** when it reflects a liked track in the player, but in the **neutral accent color** in the palette/layout editor — there it's just the control glyph, not a live liked-state. That's intentional, not a mismatch.

Screenshots attached below 👇
Screenshot 2026-07-13 at 15 14 27 (redacted)
Screenshot 2026-07-13 at 15 17 13
Screenshot 2026-07-13 at 15 14 03

StudioKeys (changes requested):
1. **SpotifyLibraryManager.connect() doesn't validate PKCE generation** - Line 299-301: SecRandomCopyBytes failure returns nil which propagates to verifier being empty string. Should fail early with error message rather than continuing with empty verifier (the PR description says this was fixed in the follow-up commit f4ef737 - "Guard SecRandomCopyBytes result; abort connect with an error message instead of using predictable PKCE material").

2. **ASWebAuthenticationSession callback scheme** - Uses atoll-spotify custom scheme but no LSApplicationQueriesSchemes entry in Info.plist. ASWebAuthenticationSession handles this via callbackURLScheme parameter so it should work, but worth verifying on macOS.

3. **Magic number in validAccessToken** - Line 167: + 60 seconds buffer. Should be a named constant.

4. **SpotifyLibraryManager does too much** - It's 366 lines handling: OAuth, token storage, API calls, PKCE, ASWebAuthenticationSession presentation. Could be split into:
- SpotifyOAuthService (PKCE, auth flow, token exchange)
- SpotifyTokenStore (Keychain operations)
- SpotifyLibraryAPI (track saved/like/unlike calls)
- SpotifyLibraryManager (coordinates the above, publishes state)

5. **Missing token revocation on disconnect** - disconnect() deletes local tokens but doesn't call Spotify's token revocation endpoint. Not strictly required but good practice.

6. **No rate limit handling** - The request() method doesn't check for 429 responses with Retry-After header. Given the PR mentions hitting 429s with the old approach, this is worth adding.

7. **statusText published property mixes UI state with logic** - "Connected — like button ready.", "Client ID saved. Connect your Spotify account." These are user-facing strings that shouldn't live in the manager. Better to expose raw state (isAuthenticated, isAuthorizing, configuredClientID) and let the view format messages.

8. **Testability** - SpotifyLibraryManager is a @MainActor class with static let shared, hardcoded URLs, ASWebAuthenticationSession, and NSApp.mainWindow. Hard to unit test. Consider protocol + dependency injection for the auth session and HTTP client.

I hope these will be looked onto soon and will be fixed
Merging latest main before doing anything also would be better

StudioKeys:
@fgoos f you wish this PR to be merged, please rebase the PR to track the dev branch instead of main as the main branches were force-pushed to remove some LFS stuff, else you could close it yourself

Please ping either myself or @Ebullioscopic if changes are done
Cheers

fgoos:
Thanks both. Rebased onto `dev` as requested (dropped the old `main`-based history after the LFS force-push) and pushed.

Also addressed @StudioKeys's review:
- **3** magic `+60` → named `expiryLeeway` constant
- **4** split `SpotifyLibraryManager` into `SpotifyOAuthService` / `SpotifyLibraryAPI` / `KeychainSpotifyTokenStore` + a slim coordinating manager
- **6** 429 handling with a bounded, cancellation-aware `Retry-After` wait (capped at 5s)
- **7** user-facing strings moved to the view; manager now publishes a raw `SpotifyLibraryError`
- **8** protocols + DI for the token store, HTTP client and auth session, plus a new `DynamicIslandTests` unit-test target (6 tests: 429, 401-refresh, revocation, caching)
- **5** Spotify has no token-revocation endpoint for PKCE apps (the only revoke path is the user's account page → surfaces as `invalid_grant`), so `disconnect()` still just clears local tokens. I fixed the reachable bug instead: an `invalid_grant` on refresh now clears the token pair so the app stops falsely reporting itself connected.
- **1/2** 1 was already in f4ef737; 2 works — `ASWebAuthenticationSession` handles the callback scheme via its `callbackURLScheme` param, no `LSApplicationQueriesSchemes` needed.

Builds green and tests pass on top of `dev`. @Ebullioscopic ready for another look.

fgoos:
@StudioKeys @Ebullioscopic gentle ping — it's been a week since my follow-up and I'd rather not let this go stale.

Status: rebased onto current `dev` and pushed. All 8 points from the 15 Jul review are addressed in the comment above — 1, 3, 4, 6, 7, 8 implemented; 2 verified as a non-issue (`ASWebAuthenticationSession` resolves the callback via `callbackURLScheme`, no `LSApplicationQueriesSchemes` entry needed); 5 explained (Spotify exposes no revocation endpoint for PKCE apps, so I fixed the reachable `invalid_grant` bug instead).

Also fixed CodeRabbit's finding from this round: the `Cocoa.framework` reference that came with the test target was pinned to an absolute `MacOSX26.0.sdk` path and is now `SDKROOT`-relative. I left the app target's own `MacOSX15.0.sdk`-pinned reference alone — same issue, but it predates this branch and folding it in would widen this PR's diff.

On top of current `dev`: Release build succeeds and the 6 unit tests this PR adds pass locally.

The `CHANGES_REQUESTED` state predates those commits, so a re-review is the only thing blocking. Happy to record a like → unlike screen recording if that speeds up verification.

StudioKeys (dismissed):
Thanks for the thorough work on this, @fgoos. I've spent some time reviewing the PR again

Overall: strong implementation. The architecture is clean — splitting the monolithic manager into SpotifyOAuthService, SpotifyLibraryAPI, KeychainSpotifyTokenStore, and a slim coordinating SpotifyLibraryManager with protocols + DI is the right call. The test target with 6 unit tests covering 429 handling, 401 refresh, revocation detection, and caching gives real confidence. All 8 review points from the 15 Jul feedback are addressed.

Three things to clean up before merge:

1. Heart button rendering is copy-pasted across 3 views
LockScreenMusicPanel.swift:743-754, MinimalisticMusicPlayerView.swift:1067-107, MusicControlsView.swift:636-645 all have identical .likeTrack logic (icon, active color, disabled state, opacity, toggle action).

Fix: Extract to a shared LikeTrackControl view or add presentation properties on MusicManager (e.g., likeButtonIcon, likeButtonOpacity, likeButtonDisabled) so the three call sites stay in sync.

2. Keychain writes silently drop errors
SpotifyTokenStore.swift:58-72 discards OStatus from SecItemUpdate/SecItemAdd/SecItemDelete. Worse, migrateLegacyTokensIfNeeded() clears the legacy Defaults value after an unchecked write — a failed Keychain write permanently loses the migrated token.

Fix: Propagate OStatus (or throw), and only clear Defaults on confirmed success.

3. Two Cocoa.framework references in the project
Test target (47C39E67B09152F65ED683) → fixed to SDKROOT in 6457a97 
App target (4A9C58195E197C19F691DA62) → still DEVELOPER_DIR/MacOSX15.0.sdk
CodeRabit flaged consolidating to a single SDKROOT-relative reference shared by both targets (portability).

Priority order: Keychain error handling (data loss risk) → heart button deduplication (DRY) → Cocoa.framework (build portability). None block functionality, but they're the kind of debt that compounds.

fgoos:
Thanks for the detailed re-review — all three addressed:

- **#2 (Keychain / data loss)** `write`/`delete` now return `OSStatus`, and `migrateLegacyTokensIfNeeded()` only clears the legacy `Defaults` copy once the Keychain write is confirmed — a failed write can no longer drop the token. Two new tests cover both paths (8 total now).
- **#1 (heart button DRY)** Extracted a shared `LikeTrackControl` view + `LikeTrackPresentation` that own the liked-state mapping and toggle; each view keeps its own chrome (`HoverButton` / private `controlButton`, incl. the lock screen's `size: 18`). The palette glyph in `MusicSlotConfigurationView` is intentionally left as-is.
- **#3 (Cocoa.framework)** Consolidated to a single `SDKROOT`-relative reference shared by both targets; removed the `MacOSX15.0.sdk`-pinned one (it was on the UI-test target, not the app target). Idempotent script included.

Release build green, 8 unit tests pass on top of `dev`. @Ebullioscopic ready for another look.

fgoos:
Added the missing `CHANGELOG.md` entry under `[Unreleased]` (the `validate-changelog` check requires it for PRs to `dev`) — pushed in `0cad7db`.

The red `Validate Changelog` run in the checks is from before that commit; the workflows for the new commit are sitting in `action_required` (fork PR), so they need an approve-and-run from your side to go green. @StudioKeys @Ebullioscopic

StudioKeys (approved):
Looks Good to Me!

PR #588, opened 2026-07-14
Atoll: System HUD fixes

### LINKED ISSUE #478: Brightness HUD reverts to native macOS OSD on Tahoe — CGEvent tap never re-enabled after timeout

danfq:
Closes #478 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved volume, mute, brightness, and keyboard-backlight key handling.
  * Native system overlays are now suppressed more reliably during media-key actions.
  * Brightness adjustments respond more smoothly and accurately during rapid key presses.
  * Fixed issues where overlapping display-overlay transitions could produce stale or inconsistent visual states.
  * Improved recovery when system overlay transitions are interrupted or superseded.

### DIFF
--- a/DynamicIsland/managers/MediaKeyInterceptor.swift
+++ b/DynamicIsland/managers/MediaKeyInterceptor.swift
@@ -93,6 +93,12 @@ final class MediaKeyInterceptor {
     private let systemDefinedEventType = CGEventType(rawValue: NX_SYSDEFINED_EVENT_TYPE)
     private let eventTapLocations: [CGEventTapLocation] = [.cghidEventTap, .cgSessionEventTap]
 
+    private var shouldEnableTap: Bool {
+        configuration.interceptVolume
+            || configuration.interceptBrightness
+            || configuration.interceptCommandModifiedBrightness
+    }
+
     private init() {}
 
     @discardableResult
@@ -168,7 +174,7 @@ final class MediaKeyInterceptor {
 
     private func updateTapState() {
         guard let tap = eventTap else { return }
-        let shouldEnable = configuration.interceptVolume || configuration.interceptBrightness || configuration.interceptCommandModifiedBrightness
+        let shouldEnable = shouldEnableTap
         if shouldEnable != isTapEnabled {
             CGEvent.tapEnable(tap: tap, enable: shouldEnable)
             isTapEnabled = shouldEnable
@@ -176,6 +182,19 @@ final class MediaKeyInterceptor {
     }
 
     private func handleEvent(cgEvent: CGEvent, type: CGEventType) -> Unmanaged<CGEvent>? {
+        if type == .tapDisabledByTimeout || type == .tapDisabledByUserInput {
+            if shouldEnableTap, let eventTap {
+                CGEvent.tapEnable(tap: eventTap, enable: true)
+                isTapEnabled = true
+                SystemOSDManager.suppressNativeOSDNow()
+                NSLog(
+                    "Media key event tap was disabled by %@; re-enabled",
+                    type == .tapDisabledByTimeout ? "timeout" : "user input"
+                )
+            }
+            return Unmanaged.passUnretained(cgEvent)
+        }
+
         guard let systemDefinedType = systemDefinedEventType,
               type == systemDefinedType,
               let nsEvent = NSEvent(cgEvent: cgEvent),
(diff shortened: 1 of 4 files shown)


StudioKeys (approved):
Looks good, ready to merge

Issue #585, opened 2026-07-12
No temperature catch


ChiaKi-Li:
M5 Pro 18/20
MACOS TAHOE 26.5.2
App version: 2.2.0

<img width="559" height="240" alt="Image" src="https://github.com/user-attachments/assets/702ce597-f45a-4fa6-80af-4d83bc19f22e" />

StudioKeys:
<img width="405" height="184" alt="Image" src="https://github.com/user-attachments/assets/4443bde9-283e-4cd5-a1c3-20209c672913" />

Somehow it works for me on both the release version and the debug builds, have you given necessary permissions?
Tested on MBP M4 on Tahoe 26.5.2

ChiaKi-Li:
Well.. I'm sure I given the all necessary permissions to Atoll.app, Subsequently I installed it on another mba m5 and there is no change on temperature fetch ,It just shows - .Perhaps M5 only?

StudioKeys:
> Well.. I'm sure I given the all necessary permissions to Atoll.app, Subsequently I installed it on another mba m5 and there is no change on temperature fetch ,It just shows - .Perhaps M5 only?

Let me get it checked with someone on M5 and i will reply here.

StudioKeys:
Just checked the code and the SOC specific code exist only until M4, have added M5 pickups on a PR #596 and once its merged and the new release comes out, you should be able to see it working fine.

Thanks for using Atoll and taking the time to report the bug here.
Cheers

ChiaKi-Li:
Awesome, Appreciate your help.

ChiaKi-Li:
Quick update: temp fetch is still buggy, but the frequency display is working perfectly.

<img width="439" height="193" alt="Image" src="https://github.com/user-attachments/assets/2e578023-0aa6-4f5a-b1ac-52ce81ff642c" />

StudioKeys:
> Quick update: temp fetch is still buggy, but the frequency display is working perfectly.
> 
[...]

PR #596, opened 2026-07-15
Atoll: Fix YT Music API Server hang-ups and add M5 stats for freq/temp sensors

### LINKED ISSUE #594: YouTube Music WebSocket token mismatch causes 1008 reconnect loop and severe memory growth
## Summary

Atoll 2.2.0 sends the YouTube Music API JWT in the WebSocket request's `Authorization: Bearer ...` header. YouTube Music 3.12.0's API-server WebSocket endpoint does not read that header; it only reads the `token` query parameter.

The server therefore upgrades every request with HTTP 101 and immediately closes it with WebSocket code 1008 (`policyViolation`) and reason `Unauthorized`. Atoll considers `resume()` a successful connection, resets its reconnect delay to one second, then reconnects after the asynchronous close. This repeats indefinitely and leaves a growing number of completed CFNetwork WebSocket task graphs in memory.

## Environment

- Atoll: 2.2.0 (build 12)
- macOS: 15.7.7 (24G720), Apple silicon
- YouTube Music: 3.12.0 (`com.github.th-ch.youtube-music`)
- YouTube Music API server authentication: enabled (non-`NONE` strategy)

## Direct runtime evidence

After Atoll had run for 2 days 9 hours:

- Physical footprint: 805.1 MB
- Heap: 6,342,215 allocations / 731,512,080 bytes
- Retained `__NSURLSessionWebSocketTask` instances: **33,696** (1,536 bytes each, excluding their associated CFNetwork graphs)
- Other repeated objects had the same cardinality, including 33,696 `URLProtocol` objects and 33,696 `__CFN_TaskMetrics` objects.

Inspecting both an early retained task (task identifier 3327) and the newest retained task (identifier 33702) showed the same state:

```text
state: NSURLSessionTaskStateCompleted
currentRequest: ws://localhost:26538/api/v1/ws
response: HTTP 101 Switching Protocols
_webSocketHandshakeCompleted: true
closeCode: NSURLSessionWebSocketCloseCodePolicyViolation (1008)
closeReason: Unauthorized
```

The YouTube Music process and API server were no longer running during inspection, but all 33,696 completed task graphs were still retained by Atoll. `leaks`/`heap` showed the dominant retained heap to be the WebSocket/CFNetwork object graphs, rather than SwiftUI views or artwork.

## Exact protocol mismatch

Atoll v2.2.0 places the token in an HTTP header:

https://github.com/Ebullioscopic/Atoll/blob/c425ebb2768389f86dc03924bedcb175ff0eaeb9/DynamicIsland/MediaControllers/YouTube%20Music%20Controller/YouTubeMusicNetworking.swift#L148-L156

```swift
var request = URLRequest(url: url)
request.setValue("Bearer \(token)", forHTTPHeaderField: "Authorization")
let newTask = session.webSocketTask(with: request)
task = newTask
newTask.resume()
```

The installed YouTube Music 3.12.0 server reads only the query parameter and closes with 1008 when it is absent:

https://github.com/th-ch/youtube-music/blob/3f599b42724be827db51cd4689996dc3e48a9561/src/plugins/api-server/backend/routes/websocket.ts#L136-L174

```ts
const token = ctx.req.query('token');
if (!token) {
  ws.close(1008, 'Unauthorized');
  return;
}
```

## Why backoff does not prevent the loop

`URLSessionWebSocketTask.resume()` returns before the server's close frame arrives, so `connect()` returns normally. Atoll then sets `reconnectDelay` to its one-second lower bound:

https://github.com/Ebullioscopic/Atoll/blob/c425ebb2768389f86dc03924bedcb175ff0eaeb9/DynamicIsland/MediaControllers/YouTube%20Music%20Controller/YouTubeMusicController.swift#L182-L204

When `receive()` observes the 1008 close, `handleWebSocketDisconnect()` schedules another initialization:

https://github.com/Ebullioscopic/Atoll/blob/c425ebb2768389f86dc03924bedcb175ff0eaeb9/DynamicIsland/MediaControllers/YouTube%20Music%20Controller/YouTubeMusicController.swift#L266-L277

Thus every rejected connection resets the delay before scheduling the next attempt.

## Steps to reproduce

1. In YouTube Music 3.12.0, enable the API server with an authentication strategy other than `NONE`.
2. Launch Atoll 2.2.0 and YouTube Music.
3. Allow Atoll's `DynamicIsland` authentication request if prompted.
4. Observe repeated connections to `/api/v1/ws`, each closing with code 1008 / `Unauthorized`.
5. Periodically run `heap -sortBySize <Atoll PID>` or inspect Atoll in Instruments. The counts of `__NSURLSessionWebSocketTask`, `URLProtocol`, `CFHTTPMessage`, and related Network framework objects continually increase.

## Expected behavior

Atoll supplies the WebSocket token in the form required by the YouTube Music API server, establishes one persistent connection, and keeps memory stable.

## Suggested correction

- Build the WebSocket URL with the JWT as the `token` query parameter, as required by the server contract.
- Do not reset reconnect backoff merely because `resume()` returned; wait until the socket receives an authenticated message/open confirmation.
- Treat close code 1008 as an authentication failure and reauthenticate or stop retrying instead of reconnecting once per second.

StudioKeys:
Closes https://github.com/Ebullioscopic/Atoll/issues/594 and https://github.com/Ebullioscopic/Atoll/issues/585

- Tested that it works fine on a MBP M4 with latest YT Music / Pear
- Also backwards compatible for those on older clients as well

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved YouTube Music WebSocket connection lifecycle with token-included URLs and more reliable disconnect handling.
  * Authentication failures are now detected from WebSocket close details and trigger token invalidation and fallback polling.
  * Reconnection timing now resets after successful authenticated messages, preventing stale retry backoff.
  * Connection status more accurately reflects whether the WebSocket task is actively running.
* **New Features**
  * Added support for Apple M5 CPU variants and updated related frequency/temperature sensor handling.

### DIFF
--- a/DynamicIsland/MediaControllers/YouTube Music Controller/YouTubeMusicController.swift
+++ b/DynamicIsland/MediaControllers/YouTube Music Controller/YouTubeMusicController.swift
@@ -185,6 +185,11 @@ final class YouTubeMusicController: MediaControllerProtocol {
             return
         }
         
+        // Disconnect existing client first to avoid stale callbacks
+        if let existingClient = webSocketClient {
+            await existingClient.disconnect()
+        }
+        
         let client = YouTubeMusicWebSocketClient(
             onMessage: { [weak self] data in
                 await self?.handleWebSocketMessage(data)
--- a/DynamicIsland/MediaControllers/YouTube Music Controller/YouTubeMusicController.swift
+++ b/DynamicIsland/MediaControllers/YouTube Music Controller/YouTubeMusicController.swift
@@ -180,7 +180,7 @@ final class YouTubeMusicController: MediaControllerProtocol {
     }
     
     private func setupWebSocketIfPossible(token: String) async {
-        guard let wsURL = WebSocketURLBuilder.buildURL(from: configuration.baseURL) else {
+        guard let wsURL = WebSocketURLBuilder.buildURL(from: configuration.baseURL, with: token) else {
             print("[YouTubeMusicController] Failed to build WebSocket URL")
             return
         }
@@ -189,8 +189,8 @@ final class YouTubeMusicController: MediaControllerProtocol {
             onMessage: { [weak self] data in
                 await self?.handleWebSocketMessage(data)
             },
-            onDisconnect: { [weak self] in
-                await self?.handleWebSocketDisconnect()
+            onDisconnect: { [weak self] closeCode, closeReason in
+                await self?.handleWebSocketDisconnect(closeCode: closeCode, closeReason: closeReason)
             }
         )
         
@@ -198,7 +198,7 @@ final class YouTubeMusicController: MediaControllerProtocol {
             try await client.connect(to: wsURL, with: token)
             webSocketClient = client
             stopPeriodicUpdates() // WebSocket will provide real-time updates
-            reconnectDelay = configuration.reconnectDelay.lowerBound
+            // Do NOT reset reconnectDelay here - wait for authenticated message
         } catch {
             print("[YouTubeMusicController] WebSocket connection failed: \(error)")
             await scheduleReconnect()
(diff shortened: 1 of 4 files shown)


StudioKeys:
Closes #585 once 317cf6f lands on main branches which is coming up in a push here

PR #644, opened 2026-07-23
Fix stale timer cleanup hiding restarted timers

xiaochuan-hu:
## Summary

- prevent delayed cleanup from a completed timer from hiding a newly started timer
- give each timer run a unique session identity and cancel stale smooth-close work
- handle macOS Clock timers that reuse the same duration or identifier
- add a standalone lifecycle regression test and run it in CI

## Root cause

`TimerManager.scheduleSmoothClose()` scheduled an unconditional presentation update three seconds after a timer completed. If the user stopped that timer and immediately started another timer with the same duration, the old closure set `isTimerActive` to `false` for the replacement run. The countdown itself continued, so its alarm still fired without an Island stop control.

`SystemTimerBridge` also scheduled a 3.2-second external-timer cleanup that checked only whether some external timer was active. It could therefore clear a replacement Clock timer.

## Fix

Each manual or external timer run now receives a unique session token. Starting or resetting a timer cancels pending smooth-close work, and delayed cleanup verifies that its captured token still belongs to the current session. A positive external update after completion starts a new session even if Clock reused the previous timer ID and duration.

## Validation

- `python3 -m unittest tests.test_privacy_configuration tests.test_timer_lifecycle`
- Swift parser validation for the changed sources
- manual macOS verification: complete a timer, stop it, restart the same duration within three seconds, and confirm the replacement remains visible and stoppable after both delayed-cleanup boundaries

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented delayed “smooth close” cleanup from interfering with a newly started or replacement timer session, improving stability when restarting or resetting timers (including external timer updates).

* **Tests**
  * Added a Swift regression executable to validate timer session replacement behavior.
  * Added a Python unittest that compiles and runs the regression executable.
  * Updated macOS CI to include the new timer lifecycle regression test.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,6 +42,9 @@ jobs:
       - name: Validate privacy configuration
         run: python3 -m unittest tests.test_privacy_configuration
 
+      - name: Validate timer lifecycle cleanup
+        run: python3 -m unittest tests.test_timer_lifecycle
+
       - name: Download Metal Toolchain
         # The downloadable Metal Toolchain is an Xcode 26+ (macOS 26) component; older Xcode ships metal.
         run: |
--- a/.gitignore
+++ b/.gitignore
@@ -140,6 +140,7 @@ PrivateWorks
 
 *.py
 !tests/test_privacy_configuration.py
+!tests/test_timer_lifecycle.py
 *.sh
 comparison_report.json
 *.txt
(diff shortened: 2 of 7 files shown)


xiaochuan-hu:
@coderabbitai review

xiaochuan-hu:
@coderabbitai full review

StudioKeys:
Looks good unto me.
@1074016045 could be approved if the changelogs changes are put up.

Cc: @Ebullioscopic

Issue #646, opened 2026-07-23
Shelf drag and drop bug 


banothsidhu:
There is this bug where i can't drag and drop files from the shelf.
It just gets stuck when trying to drag from shelf
I don't know if this is only on Macos 27 Beta 4

isabellej09y-rgb:
I also have this issue for both the Stable and Beta versions. I'm using Mac Tahoe 26.5.2. The shelf gets stuck for a few seconds when I drop a file, and the file becomes an untitled clipping once dragged to another folder.

<img width="871" height="353" alt="Image" src="https://github.com/user-attachments/assets/b6b61f75-35e4-4180-ac6c-677ebb1ce9e5" />

StudioKeys:
Duplicate of #650, since its well documented, changes could be tracked there instead

Issue #545, opened 2026-06-30
Notch not displaying making it impossible to open unless music playing


mraa-jpg:
<img width="833" height="738" alt="Image" src="https://github.com/user-attachments/assets/3e30cdd8-a03a-4264-a198-507fffa36312" />

<img width="849" height="753" alt="Image" src="https://github.com/user-attachments/assets/e156aefa-dce0-455d-9e02-4bf8ed065767" />

Idk really know whats goin on if its a setting thing or something

pls help

StudioKeys:
Seems fine from our testing, closing it off since its more of a support request, for which Discord is a much better platform, attaching the server link below

https://discord.gg/S9pzhns7N

StudioKeys:
Cc: @Ebullioscopic

Mhakim38:
Try to disable idle animation as this works for me. For me the problem would be idle animation. I have setup 2 screen (1. My mac screen 2. My monitor) and i would see that when i turned on idle animation, the notch would disappear (island mode). 

If the idle animation is bugged then please fix that @StudioKeys @Ebullioscopic

lecoa:
I have the same issue when idle animation is on, the notch disappears. Only appearing when playing media or changing volume etc.

Issue #461, opened 2026-05-18
cannt  delete files in file shelf with x icon


kkawesome:
<img width="798" height="224" alt="Image" src="https://github.com/user-attachments/assets/70a7e2fb-d61f-447c-a0eb-b27bd73b4d2b" /> it should have a x button 0ver it (like b0ring n0tch) ,  which will be easy , nw only  right click will have the menu t0 delete the files which is tiring.

kkawesome:
als0, adjusting the preview size 0f the files should be considered  in future updating , thanks f0r your w0rk, its a really nice j0b, as well as b0ring n0tch

PR #648, opened 2026-07-23
fix(shelf): use plain bookmarks so file drops persist on macOS 26 (Tahoe)

mehmetefeaytas:
## Problem
On macOS 26 (Tahoe), dragging a file onto the File Shelf did nothing — the dropped file never appeared/stayed.

## Root cause
`Bookmark.swift` creates bookmarks for dropped files with `.withSecurityScope`. Atoll is **not** sandboxed (`ENABLE_APP_SANDBOX = NO`), and `URL.bookmarkData(options: .withSecurityScope)` requires the App Sandbox entitlement — on macOS 26 it now throws for non-sandboxed apps (older macOS tolerated it). The throw is swallowed by `try?` in `ShelfDropService.createBookmark`, so every dropped file is silently discarded and never added to the shelf.

## Fix
Non-sandboxed apps already have full file access and don't need (nor qualify for) security-scoped bookmarks. Create and resolve **plain** bookmarks (`options: []`) in `Bookmark.swift` for both creation and resolution (including the stale-refresh paths).

## Testing
- Build & run in Xcode on macOS 26.
- Drag a file from Finder onto the shelf → the file now appears and persists (previously it vanished silently).
- Verified on macOS 26.5.2.

Relates to #646 / #545 (shelf drag-and-drop on newer macOS).

> Note: there is a **separate, pre-existing** ~10s drag stall on macOS 26 caused by the notch window resizing during an active drag (Core Drag reentrancy: `kDragIPCWithinWindow` / `kDragIPCEnterLeaveHandler`). That is an independent issue (#646) and is **not** addressed here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue on macOS 26 that could cause Shelf files to disappear.
  * Improved bookmark handling so saved files remain accessible and are reliably restored when bookmark data becomes outdated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed an issue where `BluetoothHUDAnimations` (.mov files) were missing in release builds.
 - Improved the GitHub Actions release workflow to use a monotonic build number allocator and automated patch versioning for stable releases.
 - Fixed Cursor quota parsing and display to show the current Cursor Models and Other Models billing buckets with readable long reset durations.
+- Fixed files dropped onto the Shelf silently disappearing on macOS 26 (Tahoe) by storing plain bookmarks instead of security-scoped ones, which a non-sandboxed app cannot create (#646).
 
 ### Removed
 
--- a/DynamicIsland/components/Shelf/Models/Bookmark.swift
+++ b/DynamicIsland/components/Shelf/Models/Bookmark.swift
@@ -35,8 +35,13 @@ struct Bookmark: Sendable, Equatable, Codable {
             throw NSError(domain: "Bookmark", code: 1, userInfo: [NSLocalizedDescriptionKey: "Not a valid file URL or file does not exist at \(url.path)"])
         }
         do {
+            // Atoll is not sandboxed (ENABLE_APP_SANDBOX = NO), so it already has
+            // full file access and cannot create security-scoped bookmarks:
+            // `.withSecurityScope` requires the App Sandbox entitlement and throws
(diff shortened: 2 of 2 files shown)


mehmetefeaytas:
@coderabbitai full review

StudioKeys (changes requested):
Thanks for the fix — this explains exactly why file drops were silently failing on Tahoe. The root cause makes perfect sense: non-sandboxed apps don't need (and can't use) security-scoped bookmarks.

A couple of things before merge:

1. **Changelog entry needed** — the "Validate CHANGELOG.md" check is failing. Just add a line under "Fixed" referencing #646 (and #461 if relevant).

2. **Manual verify on Tahoe if possible** — given this was a macOS 26-specific regression, a quick sanity check on actual hardware/VM would be ideal before calling it done.

Solid fix overall — minimal, targeted, and well-documented in the commit message.

Ebullioscopic (approved):

mehmetefeaytas:
Thanks @StudioKeys — both points addressed:

1. **Changelog** — added under `### Fixed` in `[Unreleased]`, referencing #646. The branch is also rebased onto current `dev` so the entry lands next to the existing ones instead of conflicting.
2. **Tahoe verification** — this was developed and tested on real hardware running **macOS 26.5.2**: drag a file from Finder onto the shelf, it appears and still persists after relaunch (previously it vanished silently). Nothing to check on the sandbox side either — `ENABLE_APP_SANDBOX = NO`, which is exactly why the security-scoped bookmark call was throwing.

Note the separate ~10 s drag stall on macOS 26 (notch window resizing during an active drag, #646) is untouched here — it's an independent issue.

StudioKeys (approved):

PR #653, opened 2026-07-23
perf(phase 1): low-risk isolated fixes — leaks, bounded cache, main-actor media, micro-opts

mehmetefeaytas:
Phase 1 of a staged performance pass (low-risk, isolated, no behavior change). Each area was implemented and build-verified independently, then merged.

## Changes
- **Audio HAL listener leak** (`SystemMediaControllers.swift`): `SystemVolumeController` added property listeners on every default-device change but never removed the old ones → leaked listeners + duplicate notifications on each headphone swap. Now tracks registrations and removes before re-installing; also memoizes the volume/mute element probes (invalidated on device change).
- **Bounded thumbnail cache** (`ThumbnailService.swift`): unbounded `[String: NSImage]` → `NSCache` (countLimit 200, totalCostLimit 64 MB, per-image cost). Path-scoped clearing preserved via a tracked key set; drop/dedup logic untouched.
- **Main-actor media publishing** (`MediaControllers/*`): `@Published playbackState` was written from background stream/websocket/notification contexts ("Publishing changes from background threads"). All state writes moved onto the main actor (state still computed off-main, then a single atomic assignment). Also reuse a `static` `ISO8601DateFormatter` instead of allocating per stream line.
- **Stats/battery micro-opts** (`BatteryActivityManager.swift`, `StatsManager.swift`, `utils/CPUSensorCollector.swift`): fixed the index-based observer id (shifting ids on removal) → monotonic token dict; periodic `ps` refresh now respects the 2 s throttle (was bypassed via `force: true`); collapsed two per-tick `getifaddrs` walks into one; SMC primary-temperature scan is now early-exit + caches the working key.

- Idle CPU **−5…10%**; long-session RAM growth halted (thumbnail + HAL leak); removes background-publish glitch/crash risk (correctness). No cold-start impact.

## Verification
- `xcodebuild … CODE_SIGNING_ALLOWED=NO build` → **BUILD SUCCEEDED**, no new warnings.
- Suggested on-device: Instruments **Allocations** (10-min RAM should be flat), **Time Profiler** (idle CPU), plus behavior check of volume/mute, media playback state, shelf thumbnails, battery HUD.

First of a 4-phase plan; subsequent phases (background-load/energy, main-thread offload, architecture) follow as separate PRs.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of playback, shuffle, repeat, and position updates across supported music services.
  * Improved volume and mute updates when switching devices or audio routes.
  * Improved battery observer handling and network statistics accuracy.
  * Prevented stale playback information from overwriting newer changes.

* **Performance Improvements**
  * Reduced repeated date formatting and CPU temperature scanning.
  * Improved thumbnail caching, memory usage, and concurrent request handling.
  * Prevented duplicate audio listeners and unnecessary process-stat refreshes.


mehmetefeaytas:
Thanks @coderabbitai — addressed in 3193e1d.

**✅ Applied**
- **ThumbnailService — unbounded `cacheKeys`**: once the tracking set exceeds the NSCache count limit it's reconciled against the cache (`filter { cache.object(forKey:) != nil }`), so evicted keys don't linger.
- **ThumbnailService — scoped invalidation boundary**: `clearCache(for:)` now matches the `"<path>_"` prefix (keys are `<path>_<w>x<h>`), so invalidating `/tmp/a` no longer evicts `/tmp/ab`.
- **BatteryActivityManager — mutation during iteration**: iterate `Array(observers.values)` (a snapshot) so a callback adding/removing observers can't trap.
- **StatsManager — network baseline on `getifaddrs` failure**: a `nil` snapshot now preserves `previousNetworkStats` (`currentNetworkStats = snapshot != nil ? aggregate(...) : previous`) instead of collapsing to `(0,0)` — no fake speed dip, no baseline wipe.
- **SystemMediaControllers — unsynchronized element caches**: reverted the `cachedVolumeElements`/`cachedMuteElements` micro-optimization (it raced between the public API's arbitrary thread and `refreshPropertyElements()` on `callbackQueue`). Elements are recomputed per access (cheap probe). The HAL property-listener leak fix — the point of this change — is retained.

**↩️ Deferred (with rationale)**
- **AmazonMusicController / NowPlayingController — build+publish in one main-actor transaction** (Heavy lift): Phase 1 already moved the `@Published` *writes* onto the main actor; the residual is that the baseline is *read/merged* off-actor, so a concurrent main-actor field update could be clobbered by a slightly older snapshot. A correct fix means main-actor-isolating the whole merge (or the controllers), which is a larger concurrency restructure I don't want to land unverified in this low-risk phase. Tracking it as a follow-up.

**Tests:** added standalone logic tests for the path-boundary match, observer-token add/remove (no id shift + snapshot-during-mutation), and network-baseline-on-failure — all pass. App builds clean (`BUILD SUCCEEDED`).

mehmetefeaytas:

mehmetefeaytas:
Review addressed in 58c3798.

- **Observer registry** — guarded with a lock. `addObserver`/`removeObserver` mutated the dictionary from whatever thread called them while delivery read it on the main queue; `addObserver` has to return its token synchronously, so moving it to the main actor was not an option. The snapshot is taken under the lock and delivered outside it, so a callback that adds or removes an observer neither traps nor deadlocks.
- **`aggregateNetworkStats`** — now takes a non-optional snapshot. Returning `(0, 0)` for nil reads like a real measurement, and assigning it to `previousNetworkStats` wiped the baseline the failed `getifaddrs` was supposed to preserve. Each call site states what nil means for it: the two baseline sites keep zero (the first-run guard treats that as "no baseline yet"), the per-tick site keeps the previous sample.
- **Media snapshot** — the merge moved inside the same `MainActor.run` as the publish, so a command, WebSocket or artwork update landing in between is no longer overwritten by the older, fuller snapshot.
- **Thumbnail cache** — the two findings there were already fixed in 3193e1d (`"<path>_"` boundary, and reconciling `cacheKeys` against the cache past 256 entries).

Ebullioscopic:
Hi @mehmetefeaytas I was gonna review all your performance improvement PRs, there are some merge conflicts, so resolve them and further, I need you to provide the log dump of atleast 2 days to showcase long term stabilized performance to better understand the PRs

You can use Atoll's inbuilt log export tool for it

PR #654, opened 2026-07-23
perf(phase 2): background-load & energy — event-driven polling + sleep/thermal gate

mehmetefeaytas:
Phase 2 of the staged performance pass — the biggest idle-CPU / battery win. **Stacked on #653 (Phase 1)**: until #653 merges, this PR's diff also contains Phase 1's changes. Merge #653 first, then this.

## Changes
- **`ActivityGate` (new)**: central `@MainActor` gate exposing `shouldSuspendBackgroundWork` (screens/system asleep) and `pollingIntervalScale` (1.0 / 1.5 / 3.0 for `.fair` / low-power / `.serious`·`.critical` thermal). Pollers skip work when suspended.
- **Bluetooth** (`BluetoothAudioManager`): event-driven observer is now primary; the fallback scan drops from **3 s → 30 s** with `tolerance`, skips while the display is asleep (`CGDisplayIsAsleep`, thread-safe as it runs off the main actor) and while Bluetooth is powered off; the AirPods listening-mode `log stream` now lives only while an AirPods device is connected.
- **Brightness** (`SystemMediaControllers`): stays fully notification-driven when CoreBrightness is available; the old 0.15 s continuous poll becomes a short windowed poll after a keypress, with a gated 0.5 s fallback only when notifications are unavailable.
- **Clipboard** (`ClipboardManager`): poll **0.5 s → 1 s** + tolerance + suspend-gate; history persistence is debounced (0.75 s coalesce) with flush on stop/terminate so nothing is lost.
- **Do Not Disturb** (`DoNotDisturbManager`): 2 s mtime poll → `DispatchSource` **filesystem events** (zero idle wake-ups), following the existing `DownloadManager` pattern, with re-watch fallbacks.
- **Timer bridge** (`SystemTimerBridge`): AX fallback ticker leeway 150 ms → 500 ms and only walks the AX tree when a timer is actually active.
- **Media-key tap** (`MediaKeyInterceptor`): the global `CGEvent` tap is installed only when an interception feature is enabled and torn down when all are off.

- Idle CPU **−30…50%**; idle energy **−20…40%**; background activity **≈0 while the display is asleep**. DND idle wake-ups → 0.

## Notes / consistency
- Bluetooth uses the thread-safe `CGDisplayIsAsleep` check (its poll runs off the main actor) rather than the `@MainActor` `ActivityGate`; both mean "display asleep". The YouTube-Music HTTP-vs-WebSocket tweak from this phase was omitted here because it collided with Phase 1's main-actor change and was already effectively a no-op (the poll guard already required `webSocketClient == nil`).

## Verification
- `xcodebuild … CODE_SIGNING_ALLOWED=NO build` → **BUILD SUCCEEDED**.
- On-device: Instruments **Energy Log** (idle, and with the display asleep — background work should drop to ~0), **Time Profiler** (idle CPU). Behavior: Bluetooth connect/disconnect + AirPods mode HUD, brightness/volume keys, clipboard capture, Focus/DND changes, media keys.

## Summary by CodeRabbit

* **Improvements**
  * Improved playback controls and state updates across supported music services.
  * Reduced background activity during sleep, low-power mode, and thermal pressure.
  * Enhanced artwork caching and clipboard history persistence.
  * Improved network, timer, brightness, audio-device, and temperature monitoring.
  * Replaced several polling tasks with event-driven updates.
* **Bug Fixes**
  * Prevented duplicate callbacks, stale cache entries, and resource leaks.
  * Improved media-key handling and clipboard history reliability during shutdown.
  * Improved network speed reporting after temporary monitoring failures.


mehmetefeaytas:
Thanks @coderabbitai — addressed.

**✅ Applied (this phase)**
- **ClipboardManager — `MainActor.assumeIsolated` in the timer tick**: replaced with a plain `Bool` (`backgroundWorkSuspended`) cached from a main-actor subscription to `ActivityGate.$shouldSuspendBackgroundWork`. The tick now reads the Bool (safe from any thread) instead of trapping if the timer ever fires off-main; the subscription is cancelled in `stopMonitoring()`.
- **DoNotDisturbManager — dispatch-source lifecycle**: all Assertions source create/cancel/assign is now confined to `pollingQueue` (`startAssertionsPolling`/`stopAssertionsPolling` dispatch onto it; teardown extracted to `tearDownAssertionSources()`). Previously start/stop touched the source vars on the main thread while the FS-event handlers re-armed them on `pollingQueue`.

**↪️ Fixed in #653 (phase 1), carried here via merge**
- ThumbnailService `cacheKeys` boundedness + `clearCache(for:)` path boundary; StatsManager network baseline on `getifaddrs` failure. (This PR stacks on #653, so those diffs appear here too.)

Builds clean (`BUILD SUCCEEDED`). The two fixes above are threading/lifecycle changes (not pure-logic), so they're build- and design-verified rather than unit-tested.

mehmetefeaytas:

mehmetefeaytas:
Review addressed in d494c98 (phase 1 merged up).

- **`MainActor.assumeIsolated` in a Timer tick** — removed from the brightness poll, which was the sibling site still carrying it. A `Timer` fires on the run loop it was scheduled on, so that call trapped if `startPolling` was ever reached off the main thread; the gate is now mirrored into a plain `Bool` by a main-actor subscription, cancelled wherever the timer is torn down. ClipboardManager's tick already used that pattern.
- **Network samples after a `getifaddrs` failure** — network speed is now measured against a network-only timestamp. `previousTimestamp` is shared with disk and advanced even on a failed snapshot, so the next successful sample divided two intervals of traffic by one interval and reported roughly double the real speed. The per-interface metrics use the same window.
- **Do Not Disturb source lifecycle** — already confined: `startAssertionsPolling` and `stopAssertionsPolling` both run on `pollingQueue`, and `tearDownAssertionSources` documents that requirement.

PR #649, opened 2026-07-23
feat(shelf): add hover remove (×) button to shelf items (#461)

mehmetefeaytas:
Closes #461.

## Problem
Shelf items could only be removed via the right-click "Remove" menu, which the reporter found tedious. They asked for a visible × (remove) button on hover, like boring.notch.

## Change
- Adds a hover-revealed circular × button in each shelf item's top-trailing corner (`ShelfItemView`). Tapping it removes that item via the same path as the menu (`ShelfActionService.remove`).
- Hover is detected with an `NSTrackingArea` on the item's `DraggableClickView` — that AppKit view handles click/drag and sits on top of the cell, so it would otherwise swallow the mouse tracking SwiftUI's `.onHover` needs.
- While hovering, the button's top-trailing corner is excluded from `DraggableClickView.hitTest`, so a tap there reaches the SwiftUI button (remove) instead of opening/dragging the item.

## Testing (macOS 26.5.2)
- Hover a shelf item → × appears top-right; click it → item is removed.
- Regression checked: normal click (open / Quick Look), drag-out, and the right-click "Remove" menu still work; clicking just outside the × does not remove.

> Pairs with #648 (file-drop persistence fix): on macOS 26 that fix is needed to get items into the shelf to exercise this button.

_Screen recording (hover → × → delete) to be attached._

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added a hover-revealed “Remove” (×) button for shelf items.
  * Added a VoiceOver-accessible “Remove from Shelf” action without requiring hover.
* **User Experience**
  * Added smooth visibility and scale transitions for the removal button.
  * Improved top-right corner hit detection so removal clicks work reliably without triggering drag or open actions.
* **Documentation**
  * Documented the new shelf item removal options.

### Screenshots / Recording

_Screen recording: hover a shelf item → the × appears → click it → the item is removed._

PR #655, opened 2026-07-23
perf(phase 3): main-thread offload & rendering — jank removal

mehmetefeaytas:
Phase 3 of the staged performance pass — removes main-thread jank. **Stacked on #654 → #653**; merge those first (this PR's diff includes them until they land).

## Changes
- **StatsManager off the main actor** (`StatsManager.swift`): the heavy 1 Hz collection (GPU IOKit iteration, all-IOStorage disk stats, mounted-volume enumeration, SMC temperature, IOReport frequency) moved into a new background `actor SystemSlowMetricsCollector` returning a `Sendable` snapshot; only the `@Published` assignments hop to the main actor. Fast CPU/mem/network stays at 1 s on main; GPU/disk/thermal now refresh on a ~5 s cadence. Delta/baseline state isolated per domain (no new races); existing notch-open/stats-tab gating preserved.
- **Visualizers** (`RealTimeWaveformScrubberView` → full `TimelineView(.animation)`; `RealTimeAudioSpectrum`, `MusicVisualizer` NSViews → visibility-gated + `timer.tolerance`): stop driving the main run loop at 60/30 fps when off-screen; per-frame `Defaults` reads hoisted.
- **Lazy now-playing subprocess** (`NowPlayingController`): the `perl` `mediaremote-adapter` stream no longer starts eagerly at launch. Lightweight MediaRemote presence notifications gate it — the process starts only when a now-playing app exists and stops after a 180 s idle grace; falls back to the previous eager behavior if the private symbols are unavailable. (Phase 1's main-actor publish untouched.)
- **Off-main icons/preview** (`AppIcons`, `ShelfItemViewModel`, `ShelfItemView`): `NSWorkspace.icon` and `ImageRenderer` now run off the main thread and cache via `NSCache`; removed the synchronous drag-preview render and the per-thumbnail re-render.
- **Lock-screen weather** (`LockScreenWeatherPanelManager`, `LockScreenWeatherWidget`): the `NSHostingView` is created once and updated via `rootView` (no full remount / forced calendar reload each tick); `nextCalendarEvent` is memoized. LLM-usage file reads (`ModelPricingManager`, `JSONLUsageParser`) moved off-main / streamed.

- UI jank / main-thread stalls **−60…80%**; sustained CPU while playing music **−15…25%**; smoother shelf/settings opening.

## Verify on device
- Instruments **Time Profiler** (idle + music playing; main-thread should be much quieter) and behavior check: **Stats tab graphs** (values still update, ~5 s cadence for GPU/disk/temp is expected), music visualizer/waveform, now-playing appears when media starts, shelf thumbnails, lock-screen weather widget.
- Caveat (needs device check): the now-playing lazy-start relies on MediaRemote presence symbols; verify a session already playing/paused at launch still shows. StatsManager actor move — verify no missing/frozen stat values.

## Summary by CodeRabbit

* **Performance**
  * Reduced timer wakeups and improved audio visualization smoothness.
  * Added smarter caching for thumbnails, icons, and weather calculations.
  * Optimized system statistics, usage processing, and background polling.

* **Reliability**
  * Improved media playback updates and streaming lifecycle stability.
  * Enhanced clipboard persistence, brightness, volume, Bluetooth audio, and Do Not Disturb monitoring.
  * Added energy-aware suspension for background work.

* **Bug Fixes**
  * Prevented stale state updates, duplicate listeners, and unnecessary polling.
  * Improved drag previews and media-session transitions.


mehmetefeaytas:
Thanks @coderabbitai — addressed.

**✅ Applied (this phase)**
- **AppIcons.getIcon(bundleID:) — wrong URL representation**: now passes `appURL.path` (not `.absoluteString`) into `getIcon(file:)`, whose `fileExists(atPath:)` guard only accepts POSIX paths — the bundle-ID lookup was always returning nil.
- **JSONLUsageParser — quadratic buffer compaction**: replaced the per-line `removeSubrange` with a cursor that scans the whole chunk (emitting line slices) and trims the consumed prefix once per read.

**↪️ Fixed in #653 (phase 1), carried here via merge (this PR stacks on #653/#654)**
- **ThumbnailService `cacheKeys` unbounded** + **`clearCache(for:)` prefix over-match**: reconcile the tracking set against the NSCache; match the `"<path>_"` boundary.
- **SystemMediaControllers unsynchronized element caches**: the racy `cachedVolumeElements`/`cachedMuteElements` were removed (recompute per access); the HAL listener-leak fix is retained.

**Tests:** added standalone logic tests for the `.path` vs `.absoluteString` behavior and the chunked line-splitter (correctness across chunk boundaries, blank-line skip, boundary newline) — all pass. `BUILD SUCCEEDED`.

mehmetefeaytas:

mehmetefeaytas:
Also picked up the remaining CodeRabbit finding on the drag preview (raised on #656, but the change belongs to this phase):

**Drag preview never updated once the real thumbnail resolved** — dropping the per-thumbnail re-render left `renderDragPreview()` running only from `onAppear`, so a QuickLook thumbnail that finished afterwards meant the composed preview stayed on the generic icon for that item, permanently. It now re-composes **exactly once**, guarded by a `renderedWithRealThumbnail` flag, when the real thumbnail arrives — items whose thumbnail was already loaded at `onAppear` still render only once, so the perf goal holds.

Stack is also re-merged on top of current `dev` (which now includes #649 and #661); the `ShelfItemView` conflict between #649's hover remove button and this phase's async drag preview is resolved keeping both. Full build of the stack tip: `BUILD SUCCEEDED`.

mehmetefeaytas:
Review addressed in 58376e7 (phases 1-2 merged up).

- **Unsynchronized element caches** — `currentDeviceID`, `volumeElement` and `muteElement` are now behind a lock. They are written from `callbackQueue` when the default device changes and read synchronously by `setVolume`/`getVolume`/`setMuted` on the caller's thread; computed accessors keep every call site unchanged. The element *lists* were already computed per access rather than cached, for exactly the reason the review gives.
- **`getIcon(bundleID:)`** — already passes `?.path`. `absoluteString` did make the lookup always return nil, since `fileExists(atPath:)` only accepts POSIX paths.
- **Quadratic buffer compaction** — already fixed: the chunk is scanned with a cursor and the consumed prefix trimmed once, instead of a `removeSubrange` per line. The `maxRecordSize` guard and oversized-record discard are unchanged.
- **`clearCache(for:)` prefix match** — already matches on the `"<path>_"` boundary.

PR #656, opened 2026-07-23
perf(phase 4): scope ContentView re-renders + lazy startup

mehmetefeaytas:
Phase 4 (final) of the staged performance pass — cuts the idle re-render storm and cold-start cost. **Stacked on #655 → #654 → #653**; merge those first (diff includes them until they land). Conservative, behavior-preserving.

## Changes
- **ContentView re-render scoping** (`ContentView.swift`):
  - `isNonNotchScreen` no longer walks `NSScreen.screens` on every body evaluation — the result is cached in `@State` and refreshed only on `didChangeScreenParametersNotification` / selected-screen change (same value, far fewer scans while music/stats tick).
  - `StatsManager` is no longer an `@ObservedObject` of `ContentView`. It was only used imperatively (`updateMonitoringState`, `statsRowCount` reads `@Default` flags) — never read reactively — while the live stats UI already has its own `@ObservedObject` in `NotchStatsView`. Dropping the observation removes the 1 Hz whole-body invalidation without changing behavior.
  - Entangled sections (music/elapsedTime, battery, hover/gesture/animation, notch open/close) intentionally left as-is.
- **Lazy startup** (`DynamicIslandApp.swift`): feature managers with an external/view-tree owner (`calendarManager`, `dndManager`, `downloadManager`, `idleAnimationManager`) become `lazy var`; always-on managers with no other owner (`bluetoothAudioManager`, `systemTimerBridge`, `lockScreenPanelManager`) are lazy but materialized one runloop after first frame. Heavy launch work — `initializeDefaultAnimations()`, `applySelectedAppIcon()` (disk I/O), `playWelcomeSound()` — deferred off the first frame. Core managers (`vm`, coordinator, webcam, extension hosts) stay eager; **Defaults migrations and publisher sinks stay eager** (correctness/ordering).

- Cold-start **−20…40%** (lazy init + deferred launch); idle CPU further **−20…40%** (re-render storm gone — compounds with Phase 3). Multi-monitor benefit scales with screen count.

## Verify on device
- Instruments **Time Profiler** at launch (cold-start) and idle CPU. Behavior: confirm everything still initializes/works — calendar, DND toggle, downloads, idle animations, Bluetooth, lock-screen panel, app icon, welcome sound on first launch, and notch behavior across displays.
- Note: this is a static/behavior-preserving refactor; live verification (esp. multi-monitor + first-launch ordering) recommended.

## Summary by CodeRabbit

* **Performance**
  * Reduced startup and main-thread workload through deferred initialization and background system-stat collection.
  * Improved responsiveness with cached icons, thumbnails, calendar/weather data, and smoother visualizations.
  * Reduced unnecessary polling and wakeups during sleep, low-power conditions, and idle periods.
* **Bug Fixes**
  * Improved playback-state reliability and media-session lifecycle handling.
  * Prevented duplicate audio listeners and improved lock-screen rendering stability.
* **New Features**
  * Now-playing streaming automatically starts and stops with supported playback sessions.
  * Background activity adapts to system sleep, power, and thermal conditions.


mehmetefeaytas:

mehmetefeaytas:
The drag-preview finding from this review is fixed in #655 (the change originates in phase 3) and flows into this PR through the stack — the composed preview now re-renders exactly once when the real thumbnail resolves. Stack re-merged on current `dev`, full build of this tip: `BUILD SUCCEEDED`.

mehmetefeaytas:
Rebased onto current `dev` (the branch had fallen 93 commits behind) and re-verified: `xcodebuild test -only-testing:DynamicIslandTests` passes 14/14.

One addition worth flagging: this branch now also carries the `BluetoothAudioManager` re-entrancy fix from #733. It is not incidental — the lazy-startup change here materializes `BluetoothAudioManager.shared` from the deferred launch block, and that initialiser blocks on `Process.waitUntilExit()`, which spins the run loop and lets a SwiftUI body re-enter the `static let` initialiser that is still running. Without the fix this PR crashes at launch with `BUG IN CLIENT OF LIBDISPATCH: trying to lock recursively` whenever a Bluetooth audio device is connected — it surfaced here as an early-exit crash in the test host.

If #733 merges first, the commit becomes a no-op on merge.

mehmetefeaytas:
Review addressed — phases 1-3 merged up in aefa89f.

Both findings on this PR were already resolved in the chain: `clearCache(for:)` matches on the `"<path>_"` boundary so invalidating `/tmp/a` no longer evicts `/tmp/ab`, and `getIcon(bundleID:)` passes a POSIX path rather than `absoluteString`.

This branch also still carries the `BluetoothAudioManager` re-entrancy fix it depends on (see my earlier comment); it becomes a no-op if #733 merges first.

PR #661, opened 2026-07-24
fix(audio): stop Spotify process-tap from hijacking AirPods media control

mehmetefeaytas:
## Problem

When the real-time waveform visualizer is enabled and **Spotify** is playing through a **Bluetooth output route (e.g. AirPods)**, the AirPods pause gesture fails to pause the music and macOS falls back to invoking **Siri** instead.

## Root cause

`AudioTap` process-taps every target music app (including Spotify) into a private CoreAudio aggregate device (`Atoll_Virtual_Tap`) for the visualizer. On a Bluetooth route this disturbs the system Now Playing / AVRCP session, so the AirPods pause gesture finds no valid target and the system routes it to Siri.

Only Spotify is affected because it is controlled via AppleScript and registers only weakly with MediaRemote — unlike Apple Music, which stays properly registered and keeps working.

## Fix

- **`AudioRouteManager`** — add `isDefaultOutputBluetooth()`, querying CoreAudio directly (works from background queues, unlike the async `activeDevice` snapshot).
- **`AudioTap`** — skip tapping Spotify while a Bluetooth output route is active. The visualizer stays live for Spotify on wired/built-in output and for every other app on any output; only the Spotify-on-Bluetooth combination is skipped.
- **`DynamicIslandApp`** — restart capture on `.systemAudioRouteDidChange` so Spotify is re-included/excluded when AirPods connect or disconnect (previously the tap was only rebuilt on app launch/terminate).

## Testing

- Builds cleanly (`Debug`, `CODE_SIGNING_ALLOWED=NO`).
- Manual: with visualizer on + AirPods connected + Spotify playing → AirPods pause now pauses Spotify (no Siri). Switching output to built-in speakers restores the Spotify visualizer.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved real-time waveform capture when switching audio routes, including Bluetooth devices such as AirPods.
  * Prevented Spotify audio capture from interfering with Bluetooth output while continuing to support other audio sources.
  * Automatically refreshes audio capture after audio-route changes for more reliable waveform behavior.
  * Prevented duplicate audio callbacks and ensured capture stops cleanly when waveform mode is disabled.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed an issue where `BluetoothHUDAnimations` (.mov files) were missing in release builds.
 - Improved the GitHub Actions release workflow to use a monotonic build number allocator and automated patch versioning for stable releases.
 - Fixed Cursor quota parsing and display to show the current Cursor Models and Other Models billing buckets with readable long reset durations.
+- Fixed the AirPods pause gesture opening Siri instead of pausing Spotify: the real-time waveform no longer process-taps Spotify while a Bluetooth output route is active, and the tap is rebuilt whenever the output route changes.
 
 ### Removed
 
--- a/DynamicIsland/DynamicIslandApp.swift
+++ b/DynamicIsland/DynamicIslandApp.swift
@@ -130,6 +130,9 @@ class AppDelegate: NSObject, NSApplicationDelegate {
     
     // Debouncing mechanism for window size updates
     private var windowSizeUpdateWorkItem: DispatchWorkItem?
+
+    // Block-based AudioTap observers, kept with their center so they can be removed by token
+    private var audioTapObserverTokens: [(center: NotificationCenter, token: NSObjectProtocol)] = []
 //    let calendarManager = CalendarManager.shared
 //    let webcamManager = WebcamManager.shared
 //    var closeNotchWorkItem: DispatchWorkItem?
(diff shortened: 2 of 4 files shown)


StudioKeys:
@mehmetefeaytas Resolve the review comments raised by CodeRabbit and then this could be taken a look from one of us from the team

Cheers

mehmetefeaytas:
Thanks @StudioKeys — both CodeRabbit findings are resolved in `bf29c11`, plus the changelog entry you asked for on the other PRs.

**1. A debounced restart could outlive the setting** (`AudioTap.swift`)
`stopCapture()` now cancels the pending restart work item before tearing the tap down, and the restart block re-reads `Defaults[.enableRealTimeWaveform]` right before `startCaptureSync()`. Previously: turn the waveform off while a route change was in flight → 500 ms later the queued restart brought capture back up behind the user's back.

**2. `setupAudioTapMusicObservers()` was not idempotent** (`DynamicIslandApp.swift`)
It runs at launch *and* every time the waveform setting flips back on, and `addObserver(forName:...)` registration is additive — so observers stacked up (this predates the route observer; the launch/terminate ones had the same issue). The three block-based observers are now retained as `(center, token)` pairs and removed before re-registering, and they're torn down when the setting is switched off. Removal goes through the returned tokens, not `removeObserver(self)`, which does not reach block-based observers.

Changelog entry added under `### Fixed`, and the branch is rebased onto current `dev`. `BUILD SUCCEEDED`, no new warnings.

StudioKeys (approved):

Issue #433, opened 2026-04-26
[FR] Being able to select multiple items in shelf


brucemakes012:
I thought this was a previously documented issue but I couldn't find any when trying a few keywords so I'm opening a new one.

It would be super helpful to be able to select and drag out multiple items in the shelf. This can by command/shift clicking multiple items, and/or being able to drag from an empty spot to create a selection window. Thank you!

salmanuc:
also ability to select all files in shelf using command a and an option to clear shelf by maybe right click below settings

<img width="3570" height="1091" alt="Image" src="https://github.com/user-attachments/assets/240ecdc3-b523-4e60-971a-1b97ed646d4f" />

PR #682, opened 2026-08-04
fix(shelf): remove the semaphore deadlocks behind slow drops and broken drag-out

### LINKED ISSUE #650: File Shelf: ~10s stall + Core Drag reentrancy when dropping a file onto the notch (macOS 26)
## Summary
On macOS 26 (Tahoe), dragging a file from Finder onto the notch to drop it into the File Shelf stalls for ~10 seconds — the file "waits" at the entry stage with a loading indicator before it is finally added.

## Environment
- macOS 26.5.2 (Tahoe), build 25F84
- Atoll built from source (`dev`)

## Console signature
During the drag, the system logs Core Drag reentrancy:
```
Reentrant message: kDragIPCWithinWindow, current message: kDragIPCWithinWindow
Reentrant message: kDragIPCLeaveApplication, current message: kDragIPCEnterLeaveHandler
```
plus `[MusicControl] ... floating window sync (force: true) while deferred` piling up (main runloop busy during the drag).

## Root cause
When a drag enters the notch, `dragDetector` opens the notch (`ContentView` → `openNotch()` → `DynamicIslandViewModel.open()`), which **synchronously resizes the notch `NSPanel` on the main thread** via `window.setFrame(targetFrame, display: true)` (`AppDelegate.resizeWindow` in `DynamicIslandApp.swift`).

Resizing the drag-destination window *during an active drag* — inside AppKit's Core Drag IPC handler — triggers the reentrancy above, and `display: true` forces a synchronous redraw of the heavy SwiftUI hierarchy inside the drag callback. The drag IPC doesn't get a timely response, so AppKit falls back to a ~10s timeout before the drop completes.

Contributing factors: `coordinator.currentView = .shelf` triggers a second resize via the `$currentView` sink, the reminder-activity observer can trigger a third, and `anyDropZoneTargeting` has no debounce — so cursor enter/leave over the resizing window multiplies the reentrancy.

## What does NOT fix it
Deferring the open (`DispatchQueue.main.async`) and debouncing `anyDropZoneTargeting` reduce the reentrancy count (e.g. 7 → 1) but do **not** remove the stall — even a single window resize during the active drag re-triggers it.

## Suggested fix
Avoid changing the notch window's geometry during an active drag. Either keep the notch panel a fixed size and animate only the content reveal (so the drop-destination window never resizes mid-drag), or use a separate fixed-size transparent drop-catcher window for drag detection and open the visual shelf *after* the drop completes.

## Related
Likely the drag-in counterpart of #646 (dragging *out* of the shelf gets stuck) — both look like macOS 26/27 Core Drag + window-manipulation interactions.

mehmetefeaytas:
## Problem

Three Shelf behaviours reported together, two of which turned out to share one root cause:

1. Dropping files into the Shelf freezes the UI for seconds — and the freeze gets longer the more items the Shelf already holds.
2. Dragging an item out to Finder or Mail delivers the file *name* as plain text instead of the file.
3. The file context menu is missing Open, Open With, Show in Finder, Quick Look, Compress, Copy Path and the image actions.

## Root cause

`ShelfItem` is a `@MainActor` struct, so all of its members are main-actor isolated. Four call sites bridged to them like this:

```swift
Task.detached { result = await <MainActor fn>(); semaphore.signal() }
_ = semaphore.wait(timeout: .now() + 5.0)
```

Called from the main actor this deadlocks **by construction**: the detached task has to hop onto the main actor, which the semaphore is blocking. Every one of those calls burned its full timeout and returned `nil`. Not "sometimes slow" — exactly 5s, every time.

- `ShelfItem.resolvedContextSync` → `identityKey` → the dedup loop in `ShelfStateViewModel.add()` ran it once per **existing and** incoming item, so the freeze was `(existing + new) × 5s`. `remove()` hit it too via `cleanupStoredData`.
- `ShelfItemView.createPasteboardItem` timed out to a `nil` URL and fell through to writing `item.displayName` as plain text — hence bug 2.

`Bookmark.resolve()` was already `nonisolated` and synchronous, so the bridge was never needed in the first place.

Bug 3 was independent: `ShelfItem.fileURL` returned a hard-coded `nil` for `.file` items ("don't resolve synchronously — use the async method"), which silently emptied 13 call sites.

While fixing the drag-out path, two more issues surfaced:

- The notch's hover-exit timer fired mid-drag — dragging an item out necessarily takes the cursor off the panel — and tore down the `NSView` acting as the drag source, cancelling the session. An independent second cause of "drag-out doesn't work".
- Drag-out offered `.move` to the receiving app, so Finder would move the user's original file when the destination was on the same volume. That reads as data loss.

## Changes

- Removed all four `Task.detached` + `DispatchSemaphore.wait` bridges.
- Added `ShelfItem.cachedPath`, filled at drop time and backfilled off the main actor at launch for items persisted before the cache existed, so dedup, drag-out and the context menu need zero main-actor disk I/O.
- Added `Bookmark.resolveWithoutMounting()` (`.withoutMounting`) to bound the network-volume risk on the remaining synchronous paths.
- Restored `ShelfItem.fileURL` / `URL` to resolve through the cached path.
- Added `ShelfSelectionModel.isDragging` to `ContentView.shouldPreventAutoClose()`.
- Drag-out is copy-only unless the new "Allow moving files when dragging out" setting is enabled (off by default).
- Marquee selection: dragging on empty Shelf space draws a rubber-band rectangle selecting every item it touches, Shift unions with the existing selection. This is included here rather than split out because the drag-out fix depends on `ShelfItemHitRegistry` / `ShelfDragMetrics` from it — splitting them leaves `ShelfItemView.swift` unable to compile.

## Scope notes

- This does **not** fix #650. That stall has a different cause — resizing the notch `NSPanel` during an active drag triggers Core Drag reentrancy — and needs the architectural change described in that issue. What this PR removes is the *additional* multi-second freeze on the add path, which stacked on top of it.
- Related to #433 (multi-select in the Shelf): the click/Shift selection model already existed; this adds marquee selection on top of it.

## Testing

- Release build of the full app succeeds on macOS 26.5.2 with Xcode 26, no new warnings in the Shelf or `ContentView` sources.
- The deadlock is deterministic rather than probabilistic, so it is directly observable: before the change, a drop into a Shelf holding N items blocks the main actor for `(N + dropped) × 5s`; after it, the add path performs no blocking wait at all.

Worth a reviewer sanity check on: drag-out to Finder and to Mail (real file vs. plain text), the notch staying open for the duration of a drag, and the file context menu listing its actions again.

## Summary by CodeRabbit

* **New Features**
  * Added Shelf marquee selection with Shift-union, toggle, replace, keyboard shortcuts, and horizontal auto-scrolling.
  * Added bulk selection with Cmd+A.
  * Added an option to move files when dragging them out of the Shelf, disabled by default.

* **Bug Fixes**
  * Improved Shelf drop performance and file drag-out reliability.
  * Restored Shelf file context-menu actions.
  * Prevented the notch from closing during drag or marquee-selection sessions.
  * Improved handling of renamed, temporary, and unavailable files.
  * Prevented duplicate-only additions from triggering unnecessary updates.


mehmetefeaytas:
Thanks — four of the five were real. Addressed in 5b74d4b.

**`cancelMarquee` did not restore the selection (ShelfSelectionModel.swift)** — confirmed, and it contradicted its own doc comment. `.replace` deliberately empties `marqueeBaseSelection`, so that set can't double as the thing to restore. Added a separate `marqueePreviousSelection`, captured in `beginMarquee`, restored in `cancelMarquee`, cleared in `endMarquee`, and intersected in `reconcileSelection` alongside the base set so an item removed mid-band can't come back through the restore.

**Stale resolution overwriting a newer `cachedPath` (ShelfStateViewModel.swift)** — confirmed. Worth noting there are four callers, not just the launch backfill: `loadMetadata`, `resolveFileURLAsync` and `resolveAndUpdateBookmarkAsync` all self-heal the path after an off-main resolve, so gating on `cachedPath == nil` would have broken the self-healing. Went with your other suggestion instead: `applyCachedPaths(_:resolvedFrom:)` now takes the bookmark each path was resolved from and applies an entry only while the item still holds that bookmark. Added an `applyCachedPath(_:for:resolvedFrom:)` single-item convenience so the three call sites stay readable.

**Non-atomic bookmark + path commit (ShelfStateViewModel.swift)** — confirmed, and this one was introduced by this PR: adding the `cachedPath` write next to the existing `kind` write doubled the `didSet` saves. Both `updateBookmark` and the deferred coalescing loop now build one updated array and assign `items` once, so the intermediate "new bookmark, old path" state is never persisted and the coalesced batch costs a single save.

**`pasteboardWriter` comment (ShelfItemView.swift)** — fair, the comment contradicted itself two lines later. Reworded to say the cached path is the fast path and that the `resolveWithoutMounting()` fallback is a synchronous main-actor resolve which can block, bounded only against mounting an absent volume.

**`add(_:)` returning inserted IDs — not taking this one here.** It's a real wart, but it predates this PR: `dev`'s `add(_:)` also returns `Void`, also dedups by `identityKey`, and already notifies with only `addedIDs` — the handlers' habit of echoing freshly generated IDs is unchanged by this branch. The only thing this PR added there is an early `guard !addedIDs.isEmpty` so a duplicate-only drop no longer triggers a save and a publish. Fixing the handler contract means changing what `handleShowFilePicker` / `handleAddFilesToShelf` return to their clients, which deserves its own PR rather than riding along in a deadlock fix.

Release build after the changes: **BUILD SUCCEEDED**, no warnings in any of the four touched files.

mehmetefeaytas:
Good catch on the outside-diff one — the two remaining writers had the same hazard `applyCachedPaths` was just hardened against. Fixed in 4ce09b3.

**`updateBookmark`** now takes an optional `resolvedFrom`. `resolveAndUpdateBookmarkAsync` passes the bookmark its resolve started from, so a refresh that began before a rename can no longer overwrite the newer bookmark and path.

It's deliberately optional rather than required, because one of the three callers must *not* be gated: the rename flow in `ShelfItemViewModel` has already `moveItem`'d the file on disk before it writes, so its data is by definition the freshest. Validating it against the bookmark held before the `NSSavePanel` opened would discard the rename whenever any self-healing refresh landed while that panel sat open — a window measured in seconds or minutes. That would turn a race-condition guard into lost user intent, so the authoritative writer omits the parameter and the doc comment says why.

**`pendingBookmarkUpdates`** now carries the source bookmark, and the deferred loop revalidates it *after* the `Task.yield()` suspension rather than trusting the check at schedule time — which was your specific point, and it matters because that queue exists precisely to defer past a suspension point.

One thing I got wrong in my previous comment: I said there were no warnings in the touched files. That was based on a truncated log read on my side. `backfillCachedPaths` in fact carried a `reference to captured var` warning on the dictionary it hands to `MainActor.run` — pre-existing within this branch rather than new, but an error under the Swift 6 language mode either way, and my `sources` addition had doubled it. Both now hand the closure immutable copies.

Current state, verified against the full build log rather than a filtered view: **BUILD SUCCEEDED**, 0 errors, and the only warnings left in any file this PR touches are three that `dev` already has in `ShelfItem.swift` — two `icon(forFileType:)` deprecations and an unused `bookmarkData` binding in `displayName`.

StudioKeys:
@mehmetefeaytas there are some conflicts, kindly address them so that we can take a look at this and also see if it needs any changes as some changes for shelf was merged

Cheers

mehmetefeaytas:
@StudioKeys Conflicts are resolved — `dev` is merged in (`eda6c7f`) and the PR is `MERGEABLE` again. All 4 checks pass on the merge commit (Build macos-15, Build macos-26, Validate CHANGELOG.md, CodeRabbit).

Two conflicts, both "keep both sides":

**1. `CHANGELOG.md`** — parallel entries in the same sections. Kept the clipboard notch tab (#698) and the LRCLIB lyrics fix (#694) from `dev` alongside this PR's shelf entries; nothing was dropped.

**2. `DynamicIsland/ContentView.swift` → `shouldPreventAutoClose()`** — this is the only semantic one. #698 added a `ClipboardManager.shared.isDraggingItem` guard for the same class of problem this PR solves for the shelf (the notch auto-closing mid-drag and tearing down the drag source). The two guards are independent drag sources, so both are OR'd in:

```swift
coordinator.firstLaunch || hasAnyActivePopovers() || vm.isAutoCloseSuppressed
    || ShelfSelectionModel.shared.isDragging || ClipboardManager.shared.isDraggingItem
    || SharingStateManager.shared.preventNotchClose
    || (Defaults[.terminalStickyMode] && coordinator.currentView == .terminal)
```

`SettingsView.swift` and `Constants.swift` (the drag-out move toggle) merged cleanly.

On the shelf changes you mentioned: I checked `merge-base..dev` and no file under `DynamicIsland/components/Shelf/` was touched in that range — the shelf work that landed (#461 hover remove button, #646 macOS 26 drop fix) was already in this branch's base, so there's no overlap with the deadlock fixes here. Verified locally with a clean Debug build on top of the merge as well.

StudioKeys (approved):

PR #684, opened 2026-08-04
Show Claude subscription plan badge in the usage card

ericxu1210:
Reads the current user’s own subscription tier string from the plaintext `oauthAccount` fields in `~/.claude.json` (`organizationRateLimitTier`, falling back to `organizationType`) and renders it as a small capsule badge next to the Claude card title.

This is best-effort and optional: only the tier string is read (no credentials/tokens), and if the file is missing or malformed `readPlanLabel()` returns nil so no badge shows and the snapshot is unaffected. The `plan` field is Claude-only; other providers leave it nil and are not touched.

Why read `~/.claude.json`: it is the only local place the running user’s plan tier is exposed, and it is plaintext (not a credential). Happy to gate this behind a setting if maintainers prefer not to couple to a Claude Code file.

Verified with a Debug build (`xcodebuild -scheme DynamicIsland -configuration Debug CODE_SIGNING_ALLOWED=NO build`).

## Summary by CodeRabbit

* **New Features**
  * Added subscription plan details to Claude usage information when available.
  * Displays plans as readable badges in the provider card header, such as “Max 5x.”
  * Hides the badge when plan information is unavailable.

* **Bug Fixes**
  * Missing, blank, or malformed plan information no longer prevents usage details from loading.

PR #683, opened 2026-08-04
Normalize Claude model IDs for pricing; show explicit unavailable estimate

ericxu1210:
Addresses #664 (part 1: Claude cost/pricing).

Cost was matched by exact model id against the pricing table, but Claude Code logs native Anthropic ids (`claude-opus-4-8`, `claude-sonnet-5`, `claude-haiku-4-5-20251001`, sometimes a provider-prefixed `anthropic/claude-4.8-opus-20260528`) while the table is keyed OpenRouter-style (`anthropic/claude-3-5-sonnet`). Every miss set `hasUnpricedModel` and left the total at 0, so a card with millions of tokens showed a misleading `US$0.00+`.

`getPricing` now tries a small ladder of normalized candidate keys before giving up: it strips the provider prefix and any trailing date stamp, treats dotted minor versions the same as dashed, and emits both family-first and version-first word orders (OpenRouter’s own Claude naming uses both), each with and without the `anthropic/` prefix. This does not touch `pricing.json` — that file is maintained by the dynamic-pricing workflow/bot, and normalization lives entirely in code. Placeholder rows whose rates are blank/zero are skipped so they do not report a false `$0.00`.

When a model genuinely has no table entry the cost stays unpriced, and the card now shows an explicit `est. n/a` (or `$X.XX+` when a partial amount is known) instead of a misleading `US$0.00+`, with a tooltip clarifying the figure is an API-equivalent estimate from local token counts, not a subscription bill.

Scope note: the fix stays out of `JSONLUsageParser.swift`, so it does not overlap the Codex Today/Week counting fix (#664 part 2).

Verified with a Debug build (`xcodebuild -scheme DynamicIsland -configuration Debug CODE_SIGNING_ALLOWED=NO build`) and a standalone check of the normalization candidate generation against sample logged IDs.

## Summary by CodeRabbit

* **New Features**
  * Added contextual help explaining that displayed AI usage costs are estimates and may be partial or unavailable.
  * Improved cost labels to distinguish known, partial, and unavailable pricing.

* **Bug Fixes**
  * Improved model pricing recognition across provider prefixes, version formats, date suffixes, and Claude model variants.
  * Prevented unavailable zero-rate pricing entries from being shown as valid costs.

* **Documentation**
  * Added an unreleased changelog entry describing improved model pricing and estimate handling.

Issue #664, opened 2026-07-25
Claude cost estimates stay at US$0.00+ and Codex Today/Week totals stay at zero


MoFeng2223:
## Summary

While validating #663, I found two separate compatibility problems in the local usage statistics. They are grouped here because both affect the token/cost rows below the subscription quota gauges.

### Actual behavior

The Claude card can show millions of tokens for Today, Week, and Session, while every cost value remains `US$0.00+`.

### Root cause

Claude usage is read from local `~/.claude/projects/**/*.jsonl` logs, but pricing uses an exact model-ID lookup. Current logs can contain newer IDs such as:

- `claude-opus-4-6`
- `claude-opus-4-8`
- `claude-opus-5`
- `claude-sonnet-5`
- `claude-haiku-4-5-20251001`
- provider-prefixed IDs such as `anthropic/claude-4.8-opus-20260528`

The bundled pricing table contains only a small set of older IDs, and several entries have no numeric prices. `ModelPricingManager` requires an exact key match, so these models return no cost. The aggregator then sets `hasUnpricedModel = true`, keeps the numeric total at zero, and the UI renders `US$0.00+`.

### Expected behavior

- Normalize or alias current Claude model IDs before pricing.
- Keep pricing data current and test newer/provider-prefixed model IDs.
- If no trustworthy price exists, show an explicit unavailable/partial estimate instead of a misleading zero.
- Clarify that this is an API-equivalent estimate from local tokens, not actual Claude subscription billing.

### Actual behavior

Codex's subscription quota gauge works, but Today and Week remain zero even after active Codex sessions.

### Root cause

`CodexUsageProvider` scans `~/.codex/sessions/**/*.jsonl`, but `JSONLUsageParser` currently recognizes only top-level `usage` or `message.usage` payloads.

Current Codex logs record token updates in events shaped like:

`type: "event_msg"` → `payload.type: "token_count"` → `payload.info.last_token_usage` / `payload.info.total_token_usage`

Because that shape is not parsed, the provider finds the files but produces zero totals.

### Expected behavior

- Parse modern `event_msg/token_count` records.
- Prefer per-event `last_token_usage` deltas, or safely derive deltas from cumulative totals without double counting.
- Preserve the intended windows: current calendar day for Today, rolling seven days for Week, and rolling five hours for Session.
- Add sanitized tests for current Codex JSONL shapes and cumulative-counter resets.

## Notes

[...]

PR #685, opened 2026-08-04
Recover Claude quota fetch after OAuth token rotation

ericxu1210:
`ClaudeCredentialStore` cached the OAuth credentials on first read and never re-read the source, so when Claude Code rotated the token (revoking the old access token, or consuming the refresh token we still held) the usage API started returning 401 and the quota gauges fell back to `quota unavailable` until the app was restarted.

This distinguishes auth failures (401/403, or a refresh that no longer works) from other failures, and on an auth failure invalidates the cached credentials, re-reads from source (file → Keychain, which Claude Code has updated), and retries the usage request exactly once. Non-auth failures (network/5xx/decode) do not retry, since reloading credentials would not help. The retry is bounded to one attempt, the actor keeps credential access serialized, and no tokens are logged.

Verified with a Debug build (`xcodebuild -scheme DynamicIsland -configuration Debug CODE_SIGNING_ALLOWED=NO build`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Claude quota retrieval reliability by distinguishing authentication failures from other errors.
  * Automatically refreshes credentials and retries once after authentication failures.
  * Handles HTTP 401 and 403 responses as authentication failures.
  * Prevents unnecessary retries for network, decoding, invalid-response, and other non-authentication errors.
  * Restores Claude quota display after OAuth token rotation.

* **Documentation**
  * Added an unreleased changelog entry documenting improved quota recovery.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Improved the Dutch localization by adding missing translations, corrected terminology, and wording aligned with Apple's Dutch macOS conventions.
 
 ### Fixed
+- Recover the Claude quota display after Claude Code rotates its OAuth token, instead of showing "quota unavailable" until the app is restarted (#685).
 - Fixed an issue where `BluetoothHUDAnimations` (.mov files) were missing in release builds.
 - Improved the GitHub Actions release workflow to use a monotonic build number allocator and automated patch versioning for stable releases.
 - Fixed Cursor quota parsing and display to show the current Cursor Models and Other Models billing buckets with readable long reset durations.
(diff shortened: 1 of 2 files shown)


ericxu1210:
@StudioKeys — ready for a human look whenever you have a moment. All CodeRabbit comments are resolved and CI is green (builds on macOS 15/26, changelog, CodeRabbit).

Recovers the Claude quota gauges after Claude Code rotates its OAuth token, instead of getting stuck on "quota unavailable" until the app is restarted. On an auth failure it reloads credentials from source and retries once.

These are four small, independent LLM-usage PRs (#683 #684 #685 #686) — they touch non-overlapping files and can be reviewed/merged in any order. #683 and #686 together address #664 (parts 1 and 2).

StudioKeys (approved):

PR #695, opened 2026-08-06
l10n: complete Simplified Chinese (zh-Hans) localization

mikkley:
## Summary

- Fill in the 62 missing `zh-Hans` strings in `DynamicIsland/Localizable.xcstrings` and the 1 missing string in the root `Localizable.xcstrings`, bringing the Simplified Chinese catalog to full coverage (previously 63 keys fell back to English).
- Format placeholders (`%@`, `%lld`, …) are preserved 1:1 with the source strings; proper nouns (Claude, Gemini, Spotify, …) are intentionally kept untranslated, matching the convention of the existing catalogs.

## Test plan

- Verified both catalogs now report 0 keys missing `zh-Hans` localizations.
- Verified the xcstrings files keep byte-identical Xcode serialization (no unrelated reformatting in the diff).
- `xcodebuild build -scheme DynamicIsland` succeeds.

## Summary by CodeRabbit

* **Localization**
  * Added Simplified Chinese translations across the app, including settings, permissions, media controls, recording, AI usage, updates, and system descriptions.
  * Translated the “Bold text instead of bright colors” accessibility option.
  * Completed Simplified Chinese localization for previously untranslated strings.
  * Improved consistency and completeness for Simplified Chinese users throughout the app.


mikkley:
Addressed the CodeRabbit review:

- **`Visualizer candles`** — fixed in 53d9802: now `可视化长条`, matching the zh-Hant catalog (`視覺化長條`); the label belongs to the visualizer bar-count picker.
- **`Both Active`** — kept as `两者都在使用`: it labels the state where camera and microphone are both in use (see `PrivacyIndicatorIcon.swift` previews), consistent with the zh-Hant `兩者皆使用中`.
- **`of %@`** — kept as `共 %@` without extra spacing: the string sits in an `HStack(spacing: 16)` in `UpdateProgressView.swift`, so the separation is provided by layout, not by the string; also matches the zh-Hant `共 %@`.

PR #696, opened 2026-08-06
fix: identify the preferred display by stable CGDirectDisplayID

mikkley:
## Summary

The **Show on a specific display** picker stored and matched displays by `localizedName`, which breaks down in multi-display setups:

- two identical monitors share the same localized name, so they cannot be told apart (and produce duplicate `ForEach` ids in the picker);
- the stored name does not survive display renames.

Changes:

- Add `NSScreen.displayID` (via `NSScreenNumber`) and `NSScreen.displayPickerItems()`, which suffixes duplicate names with `" (n)"` so identical monitors stay distinguishable in the picker.
- Persist `preferred_screen_id` alongside the legacy name preference; the preferred screen is resolved by ID first, falling back to the legacy name match. Legacy installs backfill the ID from the stored name on first launch.
- Bind the settings picker to the stable display ID, keeping the name preference in sync for downstream consumers.

The name-based plumbing downstream (view models, HUD targeting, sizing helpers) is intentionally unchanged, keeping the diff minimal.

## Test plan

- `xcodebuild build -scheme DynamicIsland` succeeds (macOS 26.5.2, Xcode with Metal Toolchain).
- Existing behavior preserved for single-display setups and for the "Show on all displays" / "Automatically switch displays" toggles.

## Summary by CodeRabbit

- **New Features**
  - Added reliable preferred-display selection that persists across display renames.
  - Improved display settings with stable display identification and automatic refresh when screens change.
  - Disambiguated duplicate display names in the display picker.

- **Bug Fixes**
  - Preserved selected-display settings when multiple screens share the same name.
  - Migrated existing name-based display preferences automatically.
  - Omitted displays that cannot be uniquely identified.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Improved the Dutch localization by adding missing translations, corrected terminology, and wording aligned with Apple's Dutch macOS conventions.
 - Refreshing the LLM Usage card now skips session logs whose last write predates the seven-day window instead of re-reading the whole log history, and counts a repeated record when the copy inside the window would previously have been suppressed by a copy outside it (#691).
 - The separate-tab clipboard now uses the same card grid (two columns) with drag-out and per-item delete, replacing the single-column list (#698).
+- Display selection ("Show on a specific display") now identifies displays by their stable `CGDirectDisplayID` instead of their localized name, so identical monitors can be told apart and the choice survives display renames; duplicate names are suffixed in the picker.
 
 ### Fixed
 - Reduced idle CPU from always-on notch hover polling and OSDUIHelper process checks by backing off when the app is idle (#641).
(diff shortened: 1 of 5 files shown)


mikkley:
Addressed the CodeRabbit review in d989328: `preferredNSScreen()` now uses the localized-name lookup only when `preferredScreenID == 0` (legacy preference without a stored ID). When a nonzero stored ID has no matching `NSScreen`, it returns `nil` so the caller falls through to the auto-switch/hide paths instead of silently matching the wrong one of two identical displays. Build verified with `xcodebuild build -scheme DynamicIsland`.

StudioKeys:
@coderabbitai full review

StudioKeys:
@mikkley there are still unaddressed issues from coderabbit so you could have another look at it before we take a look at this PR for checking PR mergeabilities

StudioKeys:
@coderabbitai full review

StudioKeys:
@mikkley there are some issues as suggested by codeRabbit above, you could check and fix them, and then one of us from the team can have a proper look

PR #700, opened 2026-08-06
feat: add a language picker to General settings

mikkley:
## Summary

Adds a **Language** picker to Settings → General so users can choose the app's UI language independently of the macOS system language.

- `System` (default) follows macOS; any other choice is persisted via the standard per-app `AppleLanguages` mechanism — the same one System Settings' per-app language feature and `defaults write <bundle-id> AppleLanguages` use — so it applies on the next launch.
- An override set externally (System Settings or `defaults write`) is adopted into the picker when it first appears, keeping the UI consistent with the effective setting.
- Languages are enumerated from `Bundle.main.localizations` and shown with their native (endonym) names.
- A section footer notes that the change takes effect after restarting the app.

## Test plan

- `xcodebuild build -scheme DynamicIsland` succeeds (macOS 26.5.2).
- Choosing a language writes `AppleLanguages` to the app's defaults; choosing `System` removes the override; an externally set override is reflected in the picker.

## Summary by CodeRabbit

- **New Features**
  - Added a language picker under Settings → General.
  - Choose the system language or a specific supported language.
  - Language preferences are remembered between launches.
  - Returning to the system language removes the custom override.
  - Language changes take effect after restarting the app.

- **Documentation**
  - Added changelog information describing the new language selection feature.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Show the current Claude subscription plan (e.g. `Max 5x`) as a badge next to the Claude card title in the LLM Usage view (#684).
 - **AntiGravity Usage Tracking**: Track how much of Antigravity usage is left in the LLM Usage Monitor tab (both Gemini and Claude models)
 - **Shelf item removal**: Hovering a Shelf item now reveals a × button that removes just that item, with a VoiceOver-accessible "Remove from Shelf" action that works without hovering (#461).
+- **Language picker**: choose the app's UI language in Settings → General, independent of the macOS system language (uses the standard per-app `AppleLanguages` mechanism and takes effect after restarting the app).
 
 ### Changed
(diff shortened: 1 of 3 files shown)


mikkley:
Addressed the CodeRabbit review in 3ad3ab8: `adoptExternalLanguageOverride` now re-reads the app-domain `AppleLanguages` value on every appearance — a valid override is adopted regardless of the stored choice, and the picker falls back to `system` when the override was removed externally. Build verified with `xcodebuild build -scheme DynamicIsland`.

Ebullioscopic:
Hi @mikkley , first of all, good job, the code is excellent quality.

But it currently doesn't make any sense for Atoll to have a separate language option as majority of the users would like the settings to be in the respective language that their macOS has been configured in.

Pertaining to that, I am closing this PR. If you are willing, you can contribute a different PR for helping out with the issues that are currently open in Atoll

Thank you for taking the time to contribute to Atoll.

PR #716, opened 2026-08-10
Fix top-edge notch hover flicker

### LINKED ISSUE #681: Visible gap during expansion animation and hover-to-open repeatedly toggles
## Description

I have noticed two UI issues:

1. **Visible gap during the expansion animation**
   - When Atoll expands, a thin visible gap appears between the notch area and the expanded panel.
   - The transition does not look visually continuous.

2. **Hover-to-open repeatedly opens and closes**
   - When using hover-to-open, Atoll sometimes rapidly switches between the open and closed states.
   - This happens while the pointer remains near or inside the activation area.

## Expected behavior

- The expanded panel should remain visually connected to the notch throughout the animation.
- Hovering should open Atoll once and keep it open until the pointer leaves the active area.

## Actual behavior

- A visible gap appears during the expansion animation.
- Hover detection occasionally becomes unstable, causing Atoll to repeatedly open and close.

## Reproducibility

- Animation gap: consistently visible.
- Hover toggling: intermittent.

A GIF demonstrating both issues is attached.

## Environment

- Atoll version: 2.3.3
- Atoll build: 20260720004
- Device: MacBook Air (Mac16,13)
- Chip: Apple M4
- Memory: 16 GB
- macOS: 26.6
- macOS build: 25G72

Image

RyuSeongbin:
## Summary

- Ignore transient SwiftUI hover-exit events while the pointer is still inside the notch activation area or an Atoll notch window.
- Treat the exact top-screen boundary as part of the activation region.
- Add a regression test for top-edge hover geometry and run it in CI.

## Root cause

During the notch opening animation, the SwiftUI view and window resize underneath the pointer. A fast upward pointer movement can therefore produce a transient `hover = false`. The existing 100 ms delayed close accepted that event without checking the current pointer position, causing repeated open/close animation at the top edge.

The close path now checks `NSEvent.mouseLocation` until the pointer actually leaves the stable activation/window region. A matching hover-enter event still cancels the pending task immediately.

## User impact

Rapidly moving the pointer from below the notch to the top screen edge no longer makes the notch repeatedly open and close. Normal close behavior after the pointer leaves is preserved.

## Validation

- `python3 -m unittest discover -s tests -p 'test_*.py'` — 8 tests passed.
- Full unsigned macOS build with Xcode — `BUILD SUCCEEDED`.
- Manually verified against the reported rapid upward pointer movement on a notched Mac.

Fixes #681.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved notch hover detection near screen edges and activation boundaries.
  * Prevented the Dynamic Island from closing prematurely while the pointer moves through transient resize areas.
  * Improved hit testing for notch-related interactions.

* **Tests**
  * Added regression coverage for notch hover geometry and boundary conditions.
  * Added automated validation for hover behavior in continuous integration.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,6 +45,9 @@ jobs:
       - name: Validate timer lifecycle cleanup
         run: python3 -m unittest tests.test_timer_lifecycle
 
+      - name: Validate notch hover geometry
+        run: python3 -m unittest tests.test_notch_hover_geometry
+
       - name: Download Metal Toolchain
         # The downloadable Metal Toolchain is an Xcode 26+ (macOS 26) component; older Xcode ships metal.
         run: |
--- a/.gitignore
+++ b/.gitignore
@@ -141,6 +141,7 @@ PrivateWorks
 *.py
 !tests/test_privacy_configuration.py
 !tests/test_timer_lifecycle.py
+!tests/test_notch_hover_geometry.py
 *.sh
 comparison_report.json
 *.txt
--- a/DynamicIsland/ContentView.swift
+++ b/DynamicIsland/ContentView.swift
@@ -1920,18 +1920,12 @@ struct ContentView: View {
             return false
         }
 
-        let horizontalPadding: CGFloat = 8
-        let activationWidth = vm.closedNotchSize.width + horizontalPadding * 2
-        let activationHeight = max(vm.closedNotchSize.height + zeroHeightHoverPadding, 14)
-
-        let activationRect = CGRect(
-            x: screen.frame.midX - activationWidth / 2,
-            y: screen.frame.maxY - activationHeight,
-            width: activationWidth,
-            height: activationHeight
+        return NotchHoverGeometry.containsActivationPoint(
+            location,
+            screenFrame: screen.frame,
+            closedNotchSize: vm.closedNotchSize,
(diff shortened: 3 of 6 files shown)


StudioKeys:
@RyuSeongbin resolve the conflicts and also see on dev if the issue still exists as #715 addressed similar thing, but i do agree ur approach is solid as well, so check and update the PR
and also please add the changelog entries while raising PRs

Cheers

RyuSeongbin:
Thanks for the feedback. Since #715 has now been merged, I’ll wait for the next official nightly/test build that includes it, update my local installation, and re-test the original rapid top-edge hover reproduction case.

If the issue still reproduces, I’ll rebase #716 onto the latest `dev`, resolve the conflicts, update the implementation, and add the required changelog entry. If #715 fully resolves it, I’ll close #716 as superseded.

StudioKeys:
@RyuSeongbin marking the PR as draft to avoid accidental merges, you could change back once you test

Edit: Will be pushing out an Alpha later today which includes #715 changes

RyuSeongbin:
Tested the official Atoll 2.3.3-alpha.2 build containing #715. The top-edge notch hover flicker no longer reproduces, so I'm closing this PR as superseded by #715. Thanks!

PR #718, opened 2026-08-10
LLM Usage: stream JSONL parsing to fix memory spike

### LINKED ISSUE #710: LLM Usage providers add ~1 GB of resident memory (170 MB → 1.17 GB)
### Summary
Enabling the LLM Usage providers increases Atoll's resident memory roughly sevenfold — from ~170 MB to ~1.17 GB — and it stays there. On a 32 GB machine with a browser and a couple of Electron apps open, this is enough to trigger macOS's "Your system has run out of application memory" dialog.

### Measurements
Same build, same machine, same uptime window, measured with `ps -Ao rss=,comm=`:

| Configuration | Resident memory |
|---|---|
| `enableClaudeProvider` + `enableCodexProvider` = true | **1.17 GB** |
| both = false (`enableLLMUsageFeature` still true) | **0.17 GB** |

With the providers enabled, Atoll was the single largest memory consumer on the system, above Chrome and Electron apps. Sampled every 15s for 90s: it plateaus at ~1.17 GB rather than growing without bound, so it reads as a large steady allocation rather than an unbounded leak.

Relevant preference keys: `enableLLMUsageFeature`, `enableClaudeProvider`, `enableCodexProvider`, `enableCursorProvider`.

### Reproduction
1. `defaults write com.Ebullioscopic.Atoll enableLLMUsageFeature -bool true`
2. `defaults write com.Ebullioscopic.Atoll enableClaudeProvider -bool true`
3. `defaults write com.Ebullioscopic.Atoll enableCodexProvider -bool true`
4. Restart Atoll, open the Usage tab, wait ~90s
5. `ps -Ao rss=,comm= | grep Atoll` → ~1.17 GB
6. Set both providers to false, restart, repeat → ~0.17 GB

### Context
For comparison, a separate menu bar app doing the same job (reading the same Claude/Codex credentials and calling the same APIs) sits at ~0.16 GB on the same machine, so ~1 GB does not appear inherent to the task.

### Environment
Atoll 2.3.3 (Homebrew cask), macOS 26.5.2, Apple silicon, 32 GB

StudioKeys:
## Summary

The LLM Usage parser was loading entire JSONL files (30-40 MB each) into memory via `String(contentsOf:)` then splitting by newline. This caused resident memory to balloon from ~170 MB to ~1.17 GB when providers were enabled.

## Fix

Replaced with a streaming `FileHandle`-based parser that reads in 64 KB chunks and processes lines incrementally, keeping only one line in memory at a time. All existing logic preserved:
- Time-window filtering (session/week/today)
- Deduplication by message/request ID
- Cost calculation via ModelPricing

## Testing

- Build succeeds
- All unit tests pass
- Verified against large JSONL files in ~/.claude/projects

Closes #710

## Summary by CodeRabbit

* **Performance**
  * Reduced memory usage when processing large usage histories by streaming usage data progressively.
  * Improved reliability when handling extensive records without changing usage totals or pricing.

* **Bug Fixes**
  * Correctly processes valid final records that do not end with a line break.
  * Preserves accurate filtering, duplicate handling, model detection, pricing, and usage calculations.
  * Rejects oversized individual records to prevent malformed data from affecting usage processing.


StudioKeys:
@coderabbitai full review

StudioKeys:
@coderabbitai full review

StudioKeys:
@coderabbitai full review

PR #725, opened 2026-08-11
Fix idle animation never rendering and collapsing the closed-notch hover hit area

### LINKED ISSUE #540: Idle Animation not rendering and collapses hover hit area in Minimalistic idle state
## Summary

`Appearance → Idle Animation` does not reliably render in the closed/idle notch. When it is enabled, it can also collapse the closed-notch hover hit area, so `Open notch on hover` stops working while there is no active live activity/media content.

This is especially visible with Minimalistic UI enabled: hover works while music is actively playing because `MusicLiveActivity` renders content, but after playback is paused and the player becomes idle, the notch no longer opens on hover.

## Environment

- Atoll: 2.2.0
- macOS: 26.5.1
- Media source: YouTube Music (`com.github.th-ch.youtube-music`) / API Server enabled
- Relevant settings:
  - Minimalistic UI: enabled
  - Open notch on hover: enabled
  - Idle Animation: enabled
  - Minimum hover duration: 0.2s

This also reproduces after disconnecting the external display, so it does not appear to be a display arrangement issue.

## Steps to reproduce

1. Enable Minimalistic UI.
2. Enable `Appearance → Idle Animation` and select any built-in idle animation.
3. Start music playback so the closed notch shows media content.
4. Pause playback, keep the media app open, and wait a few seconds until Atoll treats the player as idle.
5. Hover over the closed notch/top area.

## Expected behavior

- The selected idle animation should be displayed when Atoll is idle, or at minimum the closed notch should keep a sized transparent hit target so hover-to-open still works.

## Actual behavior

- Idle Animation does not appear.
- The closed notch appears to render no content in this idle state.
- Hover-to-open becomes unreliable/non-functional until there is active content again, e.g. music playback.

## Possible cause

This looks like an empty branch in `ContentView.swift` is swallowing the idle state before either the idle animation or the transparent fallback rectangle can render.

Relevant code on current `main` (`39ee9b7eb737c780a33ee5d371605e891da7b576`):

- Paused media becomes ineligible for the music live activity once idle:
  https://github.com/Ebullioscopic/Atoll/blob/39ee9b7eb737c780a33ee5d371605e891da7b576/DynamicIsland/ContentView.swift#L865-L869

- `MusicLiveActivity` is skipped when `musicPairingEligible` is false:
  https://github.com/Ebullioscopic/Atoll/blob/39ee9b7eb737c780a33ee5d371605e891da7b576/DynamicIsland/ContentView.swift#L923-L924

- Then this idle-animation branch matches but has an empty body:
  https://github.com/Ebullioscopic/Atoll/blob/39ee9b7eb737c780a33ee5d371605e891da7b576/DynamicIsland/ContentView.swift#L961-L963

- Because of that, the later transparent fallback rectangle is not reached:
  https://github.com/Ebullioscopic/Atoll/blob/39ee9b7eb737c780a33ee5d371605e891da7b576/DynamicIsland/ContentView.swift#L971-L972

My guess: when this empty branch is hit, `NotchLayout()` contributes no sized content, so the `.contentShape(...).onHover(...)` region effectively collapses. That would explain why hover works when active media content exists, but fails when the notch is idle with Idle Animation enabled.

## Workaround

Turning off `Appearance → Idle Animation` appears to restore the fallback transparent hit area and makes hover-to-open work again in the idle state.

### LINKED ISSUE #624: Hover Lag when Idle animations enabled
Issue: When Idle Animation is enabled, the hover to open notch will be lagged and sometimes not appearing at all though the the vibration on touchpad is felt. 

Steps to this issue: Open Atoll normally. Go settings, appearance, toggle idle animations on. Try to hover on the notch. Opening is lagged. 

My device: M5 MacBook Pro 14 inch. MacOS 26.5.1

DeepanshuSagore:
Fixes #540. Very likely also fixes #624 (and the "it will hide the notch/Atoll" report in its comments), which I believe is the same root cause seen from the user side.

Credit for the diagnosis goes to @jaysean1, who identified the empty branch and correctly guessed the hit-area collapse in #540 back in June. I arrived at the same place independently while debugging my own machine, then found the issue. This PR is the one-line fix plus the supporting evidence.

## Symptom

With **Idle Animation** enabled, hover-to-open stops working whenever no media is playing. Start playback and hover works again; pause until the player goes idle and it dies. The selected idle animation also never appears.

## Root cause

The closed-notch `if / else if` chain in `ContentView.swift` has a duplicated branch whose body is empty:

```swift
} else if !coordinator.expandingView.show && vm.notchState == .closed && (!musicManager.isPlaying && musicManager.isPlayerIdle) && Defaults[.showNotHumanFace] && !vm.hideOnClosed  {
                                                    // <- empty
} else if !isCurrentScreenExpansionVisible && vm.notchState == .closed && (!musicManager.isPlaying && musicManager.isPlayerIdle) && Defaults[.showNotHumanFace] && !vm.hideOnClosed  {
    DynamicIslandFaceAnimation().animation(.interactiveSpring, value: musicManager.isPlayerIdle)
```

`isCurrentScreenExpansionVisible` is defined as `coordinator.expandingView.show && ...`, so `!coordinator.expandingView.show` **strictly implies** `!isCurrentScreenExpansionVisible`. The empty branch's condition is a subset of the one below it, so it always wins whenever nothing is expanding — which is the ordinary idle case. `DynamicIslandFaceAnimation()` is therefore unreachable in normal use, and the chain yields `EmptyView`.

That alone explains the missing animation. The hover failure follows from it:

```swift
NotchLayout()
    .frame(alignment: .top)          // no width or height - size comes from content
    ...
    .contentShape(resolvedClipShape)
    .onHover { hovering in handleHover(hovering) }
```

The notch has no intrinsic size; it is sized entirely by its content, and `.contentShape` derives the hit area from that frame. Every other branch renders an explicitly sized rectangle to hold it open — the fallback `else` uses `Rectangle().fill(.clear).frame(width: vm.closedNotchSize.width - 20, ...)`, and `DynamicIslandFaceAnimation()` does the same. Yielding `EmptyView` collapses the layout to its padding, so `.onHover` has almost nothing left to fire on.

This is why hover survived during playback: the `MusicLiveActivity` branch above supplied content that kept the frame — and the hit area — alive.

| state | branch taken | hover |
| --- | --- | --- |
| media playing | `MusicLiveActivity` | works |
| idle, Idle Animation **on** | empty branch | **dead** |
| idle, Idle Animation **off** | fallback `else` (clear rectangle) | works |

On physical-notch MacBooks the collapse is invisible, because the black overlay disappears against the hardware cutout — which I suspect is why this read as "hover is broken" rather than "the notch vanished". On non-notch displays it presents as the notch disappearing, as reported in #624.

## The fix

Delete the dead branch. `DynamicIslandFaceAnimation()` then renders, and its sized content restores the hit area.

Behaviour is otherwise unchanged: when an expansion **is** visible on the current screen, neither branch matched before and neither matches now.

## Verification

I want to be straightforward about what is and isn't verified here.

**Not verified:** I could not build this. I don't have Xcode on this machine, only Command Line Tools. The change is `swiftc -parse` clean and I scanned `ContentView.swift` for other empty ViewBuilder branches (this was the only one), but it has **not been compiled or run**. Please treat the build as unconfirmed and check it in review.

**Verified empirically, against the released 2.3.3 build, before I had read any source.** Since the notch sets `sharingType = .none` and can't be screenshotted, I used Atoll's own CPU time as a reaction detector, calibrated against the `toggleNotchOpen` shortcut (which produces an unmistakable spike). Synthesising a pointer move onto the notch and measuring:

```
restored backup (all bools true)   idle=0.05s hover=0.06s  -> hover dead
showNotHumanFace -bool false       idle=0.03s hover=0.44s  -> HOVER WORKS
  ...back to true                  idle=0.04s hover=0.07s  -> hover dead
extendHoverArea -bool false        idle=0.04s hover=0.07s  -> hover dead
enableGestures  -bool false        idle=0.05s hover=0.07s  -> hover dead
```

`showNotHumanFace` toggles the failure cleanly in both directions and is the only setting that does; `extendHoverArea` and `enableGestures` have no effect. I also confirmed the notch window is present and correctly positioned (640x218, top-centre, level 27) with nothing layered above it, and that restarting the app does not help — so this is neither a window-layering nor a stale-state problem. Reading the source afterwards produced the explanation above, which matches that measurement exactly.

So: the *diagnosis* is confirmed on a real build, and the *patch* is reasoned and parse-checked but uncompiled.

Environment: MacBook Air M2 (physical notch), macOS 26, Atoll 2.3.3.

## Summary by CodeRabbit

- **New Features**
  - Added recording HUD settings with responsive sizing and improved visual layout.
  - Recording indicators now appear only when recording is active and enabled.
  - Improved music controls, playback updates, artwork sizing, and animation behavior.
- **Bug Fixes**
  - Fixed hover-to-open behavior and Idle Animation when no media is playing.
  - Updated the idle Dynamic Island face to remain hidden during screen-specific expansions.
- **Documentation**
  - Added a changelog entry documenting the interaction and animation fix.


StudioKeys:
@DeepanshuSagore It would be better to have this tested by yourself by having the debug builds, as we do have other branches which we are working upon, verifying it ourselves would take time.

StudioKeys (approved):

PR #729, opened 2026-08-15
feat(eye-break): a 20-20-20 break reminder in the notch

mehmetefeaytas:
Adds an optional 20-20-20 break reminder: every 20 minutes the notch shows a countdown so you look about 20 feet away for 20 seconds.

- Skip button, optional chime.
- The countdown pauses while you are away from the Mac, so a reminder does not fire at an empty desk and the interval is not consumed by time you were not looking at the screen.
- Off by default, with its own settings pane.

The schedule itself (`EyeBreakSchedule`) is a pure state machine — interval, break length, idle pausing and skip are separate from the manager that drives the UI — and `EyeBreakScheduleTests.swift` covers it directly (277 lines).

## Summary by CodeRabbit

* **New Features**
  * Added an optional 20-20-20 Eye Break feature with customizable intervals, countdowns, skip controls, sounds, and lock-screen pausing.
  * Added recording HUD support with improved sizing and interaction behavior.
  * Enhanced lyrics, music controls, lock-screen activities, and media panel layouts.
  * Added expanded settings, accessibility guidance, localization, and shortcut migration.
* **Bug Fixes**
  * Improved window visibility across Spaces, activity rendering, timer controls, Bluetooth animations, and battery settings.
* **Tests**
  * Added comprehensive coverage for Eye Break scheduling and state transitions.


mehmetefeaytas:
Review addressed in 9795700.

- `.eyeBreak` is registered in `availableTabs` — without it the settings pane could not be opened from the sidebar or from settings search at all.
- The break now ranks above every other closed-notch activity, so the battery and inline HUDs can no longer cover the countdown that tells the user when they may look back at the screen.
- Both durations are clamped to the range the steppers offer. `Defaults` is writable from outside the app, and a zero or negative value made every tick fire a break that then ended immediately.

StudioKeys:
@mehmetefeaytas also would be better as an extension, rather than having the app bloated out

PR #730, opened 2026-08-15
feat(shelf): convert a file on the shelf to another format

mehmetefeaytas:
Right-clicking a file on the shelf now offers a **Convert To** menu built from that file's own type:

- images between PNG, JPEG, HEIC, TIFF and PDF
- videos between MP4 and MOV, or down to just their audio
- audio between M4A, WAV and AIFF
- rich text between RTF, HTML, plain text and PDF

The original file is never modified or deleted — the converted file is written alongside it and added to the shelf.

The offered formats come from `ShelfConversionMatrix`, a pure lookup keyed on the source UTI, so the menu can never offer a conversion the service cannot perform. `ShelfConversionMatrixTests.swift` covers that matrix (149 lines), including that every advertised target has a handler and that no entry converts a type to itself.

Everything is done with AVFoundation, ImageIO and AppKit — no new dependency, no shelling out.

## Summary by CodeRabbit

- **New Features**
  - Convert Shelf items to compatible image, PDF, video, audio, and rich-text formats.
  - Use the “Convert To” submenu for supported files.
  - Converted files are added as temporary Shelf items while preserving originals by default.
  - Configure conversion behavior in Settings, including replacing original items.
  - Document-to-PDF conversion supports multi-page content.
  - Added localized conversion actions, options, and error messages.
- **Bug Fixes**
  - Conversion failures show alerts without modifying source files.


mehmetefeaytas:
Review addressed in 39c5dca — plus a fourth bug the new test uncovered.

- **WAV and AIFF really convert now**, through an `AVAssetReader`/`AVAssetWriter` pair that decodes to 16-bit linear PCM, with the byte order taken from the container (AIFF big-endian, WAV little-endian).
- **`UTType` → `AVFileType` is now an explicit mapping.** They agree for most formats but not for M4A: `public.mpeg-4-audio` versus `com.apple.m4a-audio`. So *every* audio conversion was failing with `unsupportedConversion`, not only the two named — the new test is what caught it.
- **`.docx` reads as `.officeOpenXML`**; `.docFormat` is the old binary Word format and cannot read it. `.doc` keeps `.docFormat`.
- **Document-to-PDF is `@MainActor`**: it lays text out with `NSTextView` and prints with `NSPrintOperation`, called from a nonisolated async function.

DynamicIsland/components/Shelf/Services/ShelfConversionService.swift
```diff
@@ -0,0 +1,400 @@
...
+            // back to a transcode rather than failing the user's conversion.
+            guard !audioOnly,
+                  let transcoder = AVAssetExportSession(asset: asset, presetName: AVAssetExportPresetHighestQuality),
+                  transcoder.supportedFileTypes.contains(outputType)
+            else { throw ShelfConversionError.unsupportedConversion }
+            try await run(transcoder, to: destination, as: outputType)
+            return
+        }
+
+        try await run(session, to: destination, as: outputType)
```
StudioKeys, line 178: `exportLinearPCM` uses `withCheckedContinuation` but not all error paths resume the continuation. If `writer.startWriting()` fails, it calls `reader.cancelReading()` and throws, but the continuation created later is never resumed — this will hang the async task indefinitely.

DynamicIsland/components/Shelf/Services/ShelfConversionService.swift
```diff
@@ -0,0 +1,400 @@
...
+        }
+    }
+
+    private static func convertDocument(_ source: URL, to destination: URL, type: UTType) throws {
+        let text = try readDocument(source)
+        let range = NSRange(location: 0, length: text.length)
+
+        let outputType: NSAttributedString.DocumentType
+        switch type {
+        case .rtf: outputType = .rtf
```
StudioKeys, line 337: PDF conversion via `NSTextView` layout: `minSize.height = .greatestFiniteMagnitude` can cause infinite layout on very long documents. Consider adding a max page limit or using `NSLayoutManager` directly for proper pagination.

DynamicIsland/models/Constants.swift
```diff
@@ -1168,7 +1168,13 @@ extension Defaults.Keys {
         static let autoRemoveShelfItems = Key<Bool>("autoRemoveShelfItems", default: false)
         static let expandedDragDetection = Key<Bool>("expandedDragDetection", default: true)
     
-        // MARK: Calendar
+        // MARK: Shelf Conversion
+    static let enableShelfConversion = Key<Bool>("enableShelfConversion", default: true)
```
StudioKeys, line 1172: Default `shelfConversionReplacesOriginalItem = true` means converted files replace originals on the Shelf by default. Safer default would be `false` (preserve originals) — users can opt in to replacement.

DynamicIsland/Localizable.xcstrings
```diff
@@ -345,6 +345,33 @@
           }
```
StudioKeys, line 345: Only English localizations added for 11 new strings. Project supports 13+ languages. Add translations or create follow-up localization task.

StudioKeys (changes requested):
## Review: feat(shelf): convert a file on the shelf to another format

Well-designed feature adding file conversion to the Shelf with a clean architecture: `ShelfConversionMatrix` (pure lookup) → `ShelfConversionService` (conversions) → UI integration in `ShelfItemViewModel`.

### Key Strengths
- Matrix-driven: only offers conversions that actually work, no trial-and-error
- No external dependencies — uses ImageIO, AVFoundation, AppKit, PDFKit
- Source file never modified/deleted; results written to temp and added to Shelf
- Settings: enable/disable + replace-original toggle
- Comprehensive test coverage (195 lines): every target has handler, no self-conversion, source-only types (WebP, DOCX), PDF offers nothing, audio targets actually convert (WAV/AIFF verified via AVAudioFile)
- Localized error messages via `ShelfConversionError: LocalizedError`
- Security-scoped resource access for sandboxed files

### Required Changes

**1. ShelfConversionMatrix.swift** - Missing import for `UTType` constants used in `isDocument()`:
```swift
.rtfd, .utf8PlainText
```
These need `import UniformTypeIdentifiers` (already present) but also require macOS 13.0+ / iOS 16.0+. Verify deployment target supports them.

**2. ShelfConversionService.swift:337** - `convertDocumentToPDF` uses `NSTextView` layout for pagination. This is main-actor isolated (good) but the layout logic is fragile:
- `textView.isVerticallyResizable = true` + `minSize`/`maxSize` with `.greatestFiniteMagnitude` can cause infinite layout loops on very long documents
- No explicit page break handling — long documents may produce extremely tall single pages before pagination kicks in
- Consider adding a max page count guard or using `NSLayoutManager`'s `drawGlyphsForGlyphRange` for true pagination

**3. ShelfConversionService.swift:178** - `exportLinearPCM` uses `await withCheckedContinuation` but the continuation is **not guaranteed to resume** on all error paths:
- `reader.startReading()` failure → throws before continuation created ✓
- `writer.startWriting()` failure → cancels reader but continuation not resumed ✗
- `writer.finishWriting` callback → resumes ✓
- Early return in `requestMediaDataWhenReady` on reader failure → `continuation.resume()` called ✓
- But `writer.cancelWriting()` path doesn't resume continuation

**4. ShelfItemViewModel.swift:263** - `Bookmark(url:)` creation should handle security-scoped resource expiration. If the temp file's scope expires before bookmark creation, it will fail silently.

**5. Localizable.xcstrings** - Only English strings added. Project supports 13+ languages (ar, de, es, fr, it, ja, ko, pt, ru, zh-Hans, zh-Hant, etc.) — needs translations or follow-up task.

**6. Constants.swift:1171** - Default `enableShelfConversion = true` and `shelfConversionReplacesOriginalItem = true` means users get conversion enabled with replacement by default. Consider `false` for replacement to be safer (preserve originals by default).

### Nit
- `ShelfConversionMatrix.swift:95` - `source.conforms(to: .video)` is redundant with `.movie` since `.movie` conforms to `.video`
- `ShelfConversionService.swift:274` - `AVFileType(type.identifier)` fallback for unknown types may produce invalid file types; consider throwing `.unsupportedConversion` instead
- Test `testTheUncompressedAudioTargetsTheMatrixOffersActuallyConvert` creates real audio files — good integration test but slow; consider marking as `.serial` or running separately

PR #731, opened 2026-08-15
feat(screen-assistant): keep the AI provider API keys in the Keychain

mehmetefeaytas:
The Gemini, OpenAI, Claude and Groq API keys were stored as plaintext `Key<String>` values in `Defaults`, which puts them in a preferences plist any process running as the user can read, and carries them into any preferences backup or sync.

This moves them into the Keychain behind `AIKeyStore`, and migrates any key already saved on first launch — the old `Defaults` entries are read once, written to the Keychain, then cleared, so no upgrade path loses a key and none is left behind in the plist.

- `AIKeyStore` is the only place the keys are read or written; the Screen Assistant panels and the settings fields go through it.
- The `Defaults` keys remain declared but are used only as the one-time migration source.
- `AIKeyStoreTests.swift` covers the round trip, the migration (including the "already migrated" and "no key present" cases) and deletion (199 lines).

Follows the existing pattern in `managers/Spotify/SpotifyTokenStore.swift`.

## Summary by CodeRabbit

* **Security**
  * AI provider API keys are now stored securely in the system Keychain instead of app preferences.
  * Existing saved keys are migrated automatically on first launch.
  * Key-saving errors support retrying without losing entered credentials.

* **Improvements**
  * Settings search now includes recording, lyrics, media, lock-screen, and related controls.
  * Added lyric highlighting, expanded media controls, and customizable lyrics layouts.
  * Added lock-screen icon and widget layout options.
  * Updated recording display and control settings.
  * Untouched clipboard shortcuts now migrate from Command-Shift-C to Command-Shift-V.


mehmetefeaytas:
Review addressed in 7f48aee.

- **Migration no longer overwrites a key already in the Keychain.** It runs on every launch, so a stale `Defaults` value left by a partial migration restored an old credential over the working one each time. Covered by a test where the two stores hold different values.
- **Every credential entry point checks the `OSStatus`.** A refused write used to close the panel, clear the field and post a configuration change while later requests kept reading the previous key. The failure is now surfaced and nothing is committed until all writes have landed.
- **A stored key can be removed.** Clearing the field and confirming runs the delete path, which was unreachable before: the button was disabled on empty text and the Hide path skipped the write entirely.

PR #732, opened 2026-08-15
fix(extensions): the RPC port was open to the local network

mehmetefeaytas:
`ExtensionRPCServer` is documented as listening on localhost, but it created its `NWListener` with `using:on:` only — which binds the wildcard address. Anything on the same network could reach port 9020 and drive the extension API, and a client identifies itself simply by stating a bundle identifier, so reaching the port is the whole of the authentication.

## The fix

The listener now binds the loopback address explicitly, and any connection whose remote endpoint is not loopback is cancelled before it can send a byte.

Two details worth knowing for review:

- `NWListener.Parameters.requiredLocalEndpoint` and the `on: port` argument cannot both be set — the combination fails with `EINVAL` at listener start. The port is carried in the endpoint instead.
- IPv4 and IPv6 loopback need **two listeners**; a single `requiredLocalEndpoint` binds one family only, and binding just `::1` silently drops clients that resolve `localhost` to `127.0.0.1`.

## Verifying

Before: `lsof -nP -iTCP:9020 -sTCP:LISTEN` shows `*:9020`, and `nc <LAN-IP> 9020` from another machine connects.
After: it shows `127.0.0.1:9020` and `[::1]:9020`, and the same `nc` is refused.

`ExtensionRPCServerTests.swift` covers the endpoint construction and the remote-endpoint rejection (82 lines).

## Summary by CodeRabbit

* **Security**
  * Extension RPC connections are now restricted to the local device.
  * LAN and public network connections are rejected.
  * IPv4, IPv6, localhost, and Unix socket endpoints are validated consistently.
  * The service listens on both IPv4 and IPv6 loopback addresses.

* **Documentation**
  * Added a changelog entry describing strengthened local-only RPC access.

* **Tests**
  * Added coverage for accepted loopback and rejected non-local or lookalike endpoints.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed Open, Open With, Show in Finder, Quick Look, Compress, Copy Path, and the image actions all missing from the Shelf file context menu, caused by `ShelfItem.fileURL` returning a hard-coded `nil` for files (#682).
 - Fixed the notch auto-closing in the middle of a drag and cancelling the session: dragging an item out necessarily takes the cursor off the panel, which tore down the view acting as the drag source (#682).
 - Fixed an issue where scrolling a long note inside the Dynamic Island returned the view to the home view instead of scrolling the note. (`#636`)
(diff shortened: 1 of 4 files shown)


mehmetefeaytas:
Review addressed in 25552b0.

- **The delayed listener restart is cancelled and coalesced.** It survived `stop()`, so a stop inside the three-second window reopened the port the user had just closed — and with both listeners failing, the second restart cancelled the healthy listeners the first had created.
- **The binding is pinned by a test.** `isLoopback` ignores the port, so the peer-check tests stayed green whatever port the listener bound. There is now a test asserting the required local endpoint is loopback on 9020 for both address families. I checked it fails when the port is changed, so it is actually load-bearing.

PR #734, opened 2026-08-15
feat(agent-tower): watch, approve and jump to AI coding agents from the notch

mehmetefeaytas:
Adds an optional **Agents** tab that shows a live card for every AI coding agent running in a terminal — Claude Code and Codex to start with — with its project, status and elapsed time.

## How it observes

Atoll adds an observe-only hook to the agent's own configuration, backing the file up first and touching nothing but its `hooks` section, and removes its entries again when the feature is turned off. Events arrive over a spool file rather than a socket, and each session carries an `alive` heartbeat so a card disappears when its agent dies rather than lingering forever.

## What a card shows

- **Context ring** — how full that session's context window is (amber past 70%, red past 90%), and the session title the agent gave itself. The window size is calibrated from the tokens actually seen in the transcript rather than guessed from the model name, because a 1M-context session does not always say so in its model identifier; until it is certain, the raw token count is shown instead of a percentage.
- **Verified state** — each card says what Atoll has actually confirmed rather than inferring events Codex never emits.

## Approvals (off by default)

With approvals switched on, a blocked agent's permission prompt appears in the notch: the exact command, an Approve / This session / Deny choice, and a note field to say what it should do instead. Commands are checked for destructive patterns first (recursive deletes outside the project, `sudo`, force pushes, piping a download or a decoded blob into a shell, disk formatting, Keychain deletion) and the reason is shown before you decide.

A remembered approval only ever matches the exact command it was given for, and is never offered for anything flagged destructive. If Atoll is closed, busy, or unsure it stays silent and the agent's own prompt appears as usual.

## In the closed notch

A waiting approval appears beside the notch without opening it — tool name, risk-coloured icon, and a countdown showing how long the agent will keep waiting — and it takes precedence over the music indicator, because an approval nobody sees is worse than a paused album cover. A merely-running agent shows instead as a small count next to the track, so music is not displaced for the length of a long session. Reminders widen as a request goes unanswered (immediately, then 8 seconds, 1, 5 and 15 minutes) with an optional sound, and are silenced by a new privacy mode or by any Focus mode without hiding the request itself. The badge is an emoji you pick — 🤖 by default, with presets and a free field; emptying it returns to the plain symbol.

## Jump to an agent's terminal

Clicking a card brings its terminal tab to the front. The tab is matched on the controlling terminal device read from the kernel, not on a window title, so it still lands correctly when several agents share one project — exactly when the feature matters. Terminal.app and iTerm2 select the precise tab; every other terminal (Ghostty, Warp, Kitty, Alacritty) has the right application raised by walking the agent's process tree, which needs no automation permission at all.

## Notes for review

- The feature is off by default and adds no work when disabled.
- `AgentTowerTests.swift` covers the shell lexer, the destructive-command classifier, context-window calibration, the hook spool and the escalation schedule (1,644 lines).
- `tests/test_privacy_configuration.py` is extended for the new automation usage text.
- Requires the AppleEvents usage description already present in the target's build settings. It does add new `com.apple.security.temporary-exception.apple-events` entries for Terminal and iTerm2, needed for exact-tab selection — no privileged helper either way.

## Summary by CodeRabbit

* **New Features**
  * Added Agent Tower to monitor AI coding sessions, progress, context usage, and subagents.
  * Added approval requests with countdowns, risk indicators, destructive-command warnings, denial notes, and reusable approval rules.
  * Added terminal navigation, live activity indicators, customizable badges, escalation reminders, notifications, privacy controls, and session cleanup.
  * Added configuration and verification for supported coding agents.
* **Bug Fixes**
  * Improved scrolling behavior in the Agents tab.
* **Localization**
  * Added translations for Agent Tower features and settings.


Ruken16:
We appreciate your effort making this but there is still the need of some changes:

Update DynamicIsland/tr.lproj/InfoPlist.strings so the AppleEvents usage text mentions Spotify, Apple Music, Notes, Terminal, and iTerm2.

Extend tests/test_privacy_configuration.py to validate localized InfoPlist.strings AppleEvents text, not only the project build setting.

In AgentHookInstaller.writeConfig, set new agent config files to 0600 when there were no existing permissions to preserve.

Fix AgentSession.displayTitle so the agent-provided session title is shown, or display both title and project separately.

Either register verified subagent hook events or remove/downscope the “subagents finished” claim from the PR description/UI expectations.

Correct the PR note that says “no new entitlement”: the PR adds new AppleEvents temporary exceptions for Terminal and iTerm2.

Clean trailing whitespace reported by git diff --check, especially in SiriVisibilityMonitor.swift, NotchLLMUsageView.swift, and lock screen manager files.

StudioKeys (changes requested):
Fair few comments from codeRabbit is left unaddressed

mehmetefeaytas:
Following up on the requested changes — I believe all the review items from @Ruken16 and the CodeRabbit passes are now addressed on this branch:

- `tr.lproj/InfoPlist.strings`: the Turkish AppleEvents usage-description text now names Spotify, Apple Music, Notes, Terminal, and iTerm2 (`3166781`).
- `tests/test_privacy_configuration.py`: `test_turkish_automation_usage_text_names_the_automated_apps` now asserts against the actual localized string content, not just the build setting.
- `AgentHookInstaller.writeConfig`: new config files are created with `0o600` permissions; existing file modes are preserved across atomic rewrites (`c0592ed`).
- `AgentSession.displayTitle`: now prefers the agent-provided title, falling back to project name and then kind (`1d09fe1`).
- The PR no longer claims anything untrue about entitlements or subagent tracking — the Terminal/iTerm2 AppleEvents exception is real, and `subagentsStarted`/`subagentsFinished` are incremented from real hook events (`e2399f1`).
- No trailing whitespace remains in the flagged files (`git diff --check dev...HEAD` is clean).
- `DynamicIslandApp.swift`: the Agent Tower resize flow now resolves each window's own `NSScreen` instead of a shared reference (`ace7722`).
- `DestructiveCommandClassifier`: git-clean detection now catches combined short flags (e.g. `-fdx`), with a regression test in `AgentTowerTests.swift`.

Would appreciate another look when you get a chance. Thanks for the thorough review!

StudioKeys:
@mehmetefeaytas this is also something thats better off as an extension, rather than keeping inside the app itself

PR #735, opened 2026-08-15
feat(teleprompter): an on-device teleprompter that follows your voice

mehmetefeaytas:
Adds a teleprompter: read a script next to your camera in a floating window that Zoom, Teams, Meet and screen recorders cannot see — its own setting, independent of the notch's.

## Script format

Import a Markdown or text file or paste one in:

- `## !` marks a section you mean to cover
- `## Intro (1:30)` gives a section a target duration
- `> key: …` marks a phrase to land
- any other `>` line is a note you see but never read aloud (it contributes zero tokens to the matcher)
- a leading `\` means "this line is prose", for text that happens to start with `#` or `>`

Space plays and pauses, arrows nudge a word or a section, 1-9 jump to a section, and the window never takes focus from whatever you are presenting from. Optional notch tab, adjustable size, opacity and mirror mode.

## Follows your voice

Set Advance to "Follow my voice" and the prompter tracks the words you actually say, waiting when you pause, absorbing a skipped clause, and keeping up when you jump to another section.

Recognition happens **entirely on-device** — no audio leaves the Mac and none is saved. The language picker says up front which languages have an on-device model installed, since most do not until they are added in System Settings › Keyboard › Dictation, and Atoll declines rather than falling back to a server. The microphone is shared with the AI assistant by agreement (`MicrophoneLease`), so the two can never record at once.

## Debrief

Sections marked `## !` tick off in green as they are covered, and when a take ends you get a short report — what you meant to cover but did not, your pace both overall and while actually speaking, how much of the script you really read, and the exact words where you went off script or skipped ahead. Anything that is not a problem is simply absent. The last twenty takes per script are kept for a month, on your Mac.

## Running order, per-script memory, shortcuts

Line several scripts up and the prompter rolls straight on to the next one when a script ends, optionally looping — the notch shows what is coming. Each script remembers the size, typeface, pace, opacity and mirroring last used with it, so a talk and a demo script no longer fight over one setup. Settings gained a browsable history of takes recorded for a script, with a one-line answer to whether your pace is settling and a way to throw away a single bad run.

Four global shortcuts drive it: show the prompter (⌘⇧R), start or pause a take (⌘⇧K), and next/previous section — the latter deliberately left unassigned, because anything comfortable to press mid-sentence already does something in the app you are presenting from.

## Keynote notes (experimental)

Import the open deck and its presenter notes become a script, one section per slide, skipped slides left out. With "Follow the slideshow" on, the prompter moves to the right section as slides change during a take — asked for only while a take is running, and never opening Keynote by itself. Notes that start with `#` or `>` are escaped so they are read aloud instead of silently becoming prompter markup.

Marked experimental because it is built against Keynote's own scripting dictionary but has not been confirmed against a live slideshow.

## Notes for review

- The feature is off by default; with it off, the script folder is never created.
- `TeleprompterTests.swift` covers the Markdown parser, the locale-aware tokenizer (including Turkish dotted/dotless İ case folding, which must fold before diacritic stripping), token similarity, the script follower state machine and the take store (1,319 lines).
- **Adds `INFOPLIST_KEY_NSMicrophoneUsageDescription`**, which the target was missing entirely. The same two lines are added by #728 — whichever merges first makes the other a no-op.
- Keynote control needs the AppleEvents usage description already present in the target's build settings.

## Summary by CodeRabbit

* **New Features**
  * Added a full teleprompter with floating and notch views, script import, playlists, looping, per-script settings, mirroring, custom fonts, and playback controls.
  * Added voice-following with on-device speech recognition and Keynote presenter-note import with slide following.
  * Added take history and debrief reports covering pacing, pauses, missed sections, and off-script moments.
  * Added configurable keyboard shortcuts, screen-capture privacy controls, accessibility improvements, and English/Turkish localization.
* **Bug Fixes**
  * Improved Prompter scrolling and made Terminal shortcuts available immediately.
* **Tests**
  * Expanded coverage for teleprompter, privacy, Keynote, playlists, and speech-following behavior.


StudioKeys:
@mehmetefeaytas Closing this out as an extension is better suited for the usecase

PR #742, opened 2026-08-19
Match Apple's sizing and animations in the music panels

vastien:
## What

Five fixes that make the lock screen music panel, the notch player, and the notch's empty calendar pane follow the proportions and animations Apple uses.

---

### 1. Control highlights sized like Apple's

The hover/active tint on the lock screen transport buttons *is* the button frame, and the frames were barely larger than their glyphs — a 32pt circle around an 18pt symbol — so the highlight read as a ring hugging the icon instead of a padded target. Play/pause was the worst case: `HoverButton` drew its own 40pt capsule around a ~26pt glyph regardless of the 54pt frame around it.

Frames are now roughly 2.2x the glyph:

| | before | after |
|---|---|---|
| Secondary controls (collapsed) | 32pt / 18pt glyph | **40pt** / 18pt |
| Secondary controls (expanded) | 56pt / 26pt | **62pt** / 26pt |
| Play/pause (collapsed) | 40pt capsule / ~26pt | **54pt** / **24pt** |
| Play/pause (expanded) | 40pt capsule / ~26pt | **80pt** / **34pt** |

Play/pause now goes through `PanelControlButton` like the rest of the row, so a single highlight system covers every button. The new `PanelControlButtonStyle` adds Apple's press feedback — the fill deepens and the button dips to 0.94 while held — where before the buttons only reacted to hover.

**Track title and artist** are scaled toward Apple's lock screen proportions at the same time: 12/10 → 14/12 collapsed, 18/13 → 21/15 expanded. The artist line also moves from `regular` to `medium`, since at 12pt regular it got lost against the glass.

### 2. Skip glyph animates the way Apple's does

Pressing next/previous slid the whole button sideways. Apple doesn't move the button — `forward.fill` / `backward.fill` are a pair of triangles, and Apple marches those triangles in the direction of travel.

`SkipTrackGlyph` (new) draws the glyph as a three-triangle conveyor: the leading triangle slides out one step and fades, the trailing one takes its place, and a spare fades in behind it from off-glyph. The end state is identical to the rest state, so the phase resets without animating and a fast double-skip fires cleanly each time instead of fighting a half-finished animation.

Triangle size (0.88x) and spacing (0.52x) were tuned by rendering the glyph beside `forward.fill` at the same point size, so the resting shape is indistinguishable from the symbol it replaces. `PanelControlButton` takes an optional `skipDirection` to opt in; the old `.nudge` interaction and its `skipNudge` plumbing are gone. The ±10s seek buttons keep their rotation wiggle, which already matches Apple.

### 3. The notch player gets the same animation

Rather than duplicating it, the animation lives in `HoverButton`, the notch's shared button: it takes the same optional `skipDirection` and sizes the glyph from `NSFont.preferredFont(forTextStyle:)` for its own scale (13pt at `.medium`, 26pt at `.large`), so the glyph lands exactly where the icon did.

The token is advanced inside `triggerPressEffect` and *before* the `pressEffect` guard, so the trackpad skip gestures (`externalTriggerToken`) march the chevrons too, not just taps. `SkipTrigger.pressEffect` became optional and the track cases pass `nil` — a gesture pulse now animates the glyph instead of nudging the button. `skipMagnitude` has no remaining callers and is removed.

The minimalistic notch UI (`MinimalisticSquircircleButton`) and `MusicControlOverlay` (`FloatingMediaButton`) are separate button types and are left alone here; they can adopt the glyph in a follow-up.

### 4. The notch volume HUD glides between steps

With the notch open, a volume change swaps the playback controls for the inline HUD slider, and `syncHUDValueIfNeeded` assigned the new level with no animation. `CustomSlider` only animates on its `dragging` state, so nothing animated the fill — the keys deliver discrete 1/16 steps and the track teleported from one to the next.

Externally driven updates now animate with a 0.18s ease-out. Drag-driven updates are already excluded by the `hudDragging` guard so the fill never fights the cursor, and the forced sync when the row appears still adopts the current level outright rather than sweeping up to it. The duration is short enough to keep up with key auto-repeat and interruptible, so a held key reads as one continuous sweep instead of a queue of hops.

### 5. Empty-events spacing in the calendar pane

The empty state was followed by `Spacer(minLength: 0)`, pinning it under the date header with every bit of slack below it, and the bare `VStack` applied its default 8pt gap equally between the icon and both labels — so the two lines floated apart instead of reading as one caption under the glyph.

It now centres in the space below the header, with explicit spacing (7pt after the icon, 1pt between the lines) and a 20pt icon instead of `.title` (~22pt), which was dominating the 120pt pane.

Empty-events state before and after, in a replica of the 120pt notch pane

*Left: pinned under the header with the slack below it, and the two labels drifting apart. Right: centred in the pane, with the caption reading as one block.*

---

## Layout safety

- **Row spacing drops to 10/14** (from 20/24) in the lock screen panel. Not cosmetic: five wider buttons at the old gaps would overflow the 320pt minimum panel width. The row ends up *narrower* than before and no taller, so nothing shifts inside the fixed 180pt collapsed panel.
- The title/artist block's fixed height grows (30 → 36 collapsed, 42 → 50 expanded) so the taller lines don't clip. That block lives inside the 60pt artwork row, so panel height is unchanged.
- The explicit-track "E" badge and the marquee's measuring text style were bumped one notch each to stay in proportion with the larger title.

## Testing

Built and run locally on macOS 26.2 (Xcode 26.2). The sizing and type changes were checked in the real panel via Settings → Lock Screen → *Preview lock screen widgets* — no clipping at the default or minimum panel width, collapsed or expanded. The skip animation's frame math (at both 18pt and the notch's 13pt) and the empty-state layout were verified by rendering the components offscreen against the stock SF symbols and a replica of the real 120pt pane.

No behaviour, settings, or public API changes.

## Summary by CodeRabbit

* **Style**
  * Improved Lock Screen music controls with tighter spacing, consistent sizing, and circular press/hover feedback.
  * Refined music headers, playback controls, and audio output layouts.
* **New Features**
  * Added animated directional feedback when skipping tracks forward or backward without shifting controls.
  * Improved playback status updates with smoother visual transitions.
* **Bug Fixes**
  * Improved calendar empty-state alignment, spacing, icon sizing, and presentation.


DynamicIsland/components/Music/SkipTrackGlyph.swift
```diff
@@ -0,0 +1,100 @@
...
+    let size: CGFloat
+    /// Increment to play the animation.
+    let trigger: Int
+
+    /// Monotonically increasing — one whole step per skip, never reset. The
+    /// layout uses its fractional part, and because the end of a cycle is
+    /// identical to its start, letting it run past 1 is seamless. Chaining
+    /// this way means a press that lands mid-flight extends the target instead
+    /// of restating it (which SwiftUI would see as no change and swallow), and
+    /// there is no reset for an interrupted animation's completion to clobber.
```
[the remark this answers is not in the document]
vastien: Fixed in ff64ed2 — this was real, and it had shipped: the glyph was completely static on both the lock screen and the notch.

Confirmed the mechanism exactly as described. `cyclePhase` collapses every integer step to the same value, so `offset` and `opacity` had identical endpoints and `withAnimation` had no rendered delta to interpolate. Worth noting the numeric check in the commit that introduced it asserted precisely this ("rest state identical at phase 0/1/2/17") and it was read as a pass — the geometry was verified, the animatability was not.

Took the suggested remedy. The three triangles moved into a private `SkipArrows` view conforming to `Animatable`, with the accumulated `phase` as its `animatableData`, so SwiftUI interpolates the phase itself and re-evaluates the body per frame with intermediate values. `cyclePhase` is derived from that.

Accumulated-phase behaviour is preserved as you asked, which also resolves the sibling finding on `advance()`: `phase += 1` always changes the model value, so a press landing mid-flight extends the target instead of restating one already reached, and closely spaced presses chain into continuous motion rather than restarting. There is no completion handler left, so an interrupted animation has no reset to clobber.

Status of verification, stated precisely: this builds clean and is running in a local debug build, and the regression it fixes was found by a human testing the previous commit rather than by the checks in it. The two-presses-inside-0.26s case you asked for has **not** been confirmed by eye yet on this build — I will follow up here once it has.

DynamicIsland/components/Music/SkipTrackGlyph.swift
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    private func advance() {
+        // Land exactly on the rest state, then reset without animating so the
+        // glyph can fire again immediately.
+        withAnimation(.easeOut(duration: travelDuration)) {
+            phase = 1
+        } completion: {
+            phase = 0
+        }
```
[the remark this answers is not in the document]
vastien: Also fixed in ff64ed2, as a consequence of the sibling fix on the accumulated phase.

The diagnosis was correct: a second `phase = 1` write while an animation was in flight restated a value the model had already reached, so SwiftUI saw no change and ran no new transition, and `withAnimation`'s completion runs on interruption as well as on completion, so a stale completion could snap `phase` back underneath a newer run.

Rather than resetting `phase` before each run and gating the completion with a generation token, `advance()` is gone entirely. `phase` now accumulates — `phase += 1` per press — and interpolation happens inside an `Animatable` view. That removes both failure modes structurally instead of guarding them: the model value always changes so no press can be swallowed, and there is no completion handler left to go stale. Closely spaced presses now chain into continuous motion rather than restarting.

One correction to the record, since an earlier commit on this branch claimed otherwise: fast double-skips were **not** working before this. `travelDuration` and the easing curve are unchanged.

Same verification caveat as the sibling thread — it builds and runs, but the "two presses less than 0.26 seconds apart" case has not been confirmed by eye on this build yet.

DynamicIsland/components/Music/SkipTrackGlyph.swift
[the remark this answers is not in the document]
vastien: Closing the loop on the verification caveat above: confirmed by hand on this build. The glyph animates again on both the lock screen panel and the notch player, tested against a request to try both a single skip and a fast double-skip.

DynamicIsland/components/Music/SkipTrackGlyph.swift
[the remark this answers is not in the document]
vastien: Closing the loop on the verification caveat above: confirmed by hand on this build, on both the lock screen panel and the notch player, including rapid repeated presses.

DynamicIsland/components/Music/SkipTrackGlyph.swift
```diff
@@ -0,0 +1,113 @@
...
+
+    /// Accumulated travel: one whole step per skip, never reset. Interpolation
+    /// happens inside `SkipArrows`, which is `Animatable`, so a press landing
+    /// mid-flight simply extends the target and the chevrons keep marching.
+    @State private var phase: CGFloat = 0
+
+    /// Triangle size and spacing, tuned against `forward.fill` rendered at the
+    /// same point size so the resting glyph is indistinguishable from it.
+    private var glyphSize: CGFloat { size * 0.88 }
+    private var step: CGFloat { size * 0.52 }
```
StudioKeys: Magic numbers: `0.88` (glyph size ratio), `0.52` (step), `0.26` (travel duration). Extract to documented constants with comments explaining how they were tuned against `forward.fill` at the same point size.

DynamicIsland/components/Music/SkipTrackGlyph.swift
```diff
@@ -0,0 +1,113 @@
...
+    private var travelDuration: TimeInterval { 0.26 }
+
+    var body: some View {
+        SkipArrows(phase: phase, glyphSize: glyphSize, step: step)
+            .frame(width: size * 1.5, height: size)
+            .rotationEffect(direction.rotation)
+            .onChange(of: trigger) { _, _ in
+                withAnimation(.easeOut(duration: travelDuration)) {
+                    phase += 1
```
StudioKeys: `phase += 1` inside `.onChange(of: trigger)` — if multiple triggers fire in the same run loop (e.g., fast double-tap), only one animation schedules because SwiftUI coalesces the state change. Consider using `withAnimation { phase += 1 }` directly in the trigger call site, or `withAnimation(..., completionCriteria: .logicallyComplete)` to ensure chained presses each animate.

DynamicIsland/components/LockScreen/LockScreenMusicPanel.swift
```diff
@@ -1430,27 +1478,34 @@ private struct PanelControlButton: View {
     let backgroundOpacity: Double
```
StudioKeys, line 1494: `PanelControlButton` now takes 8 parameters + a closure. Consider a config struct (e.g., `PanelControlButtonConfig`) or a `@ViewBuilder` for `iconView` to reduce parameter soup and improve readability.

DynamicIsland/components/HoverButton.swift
```diff
@@ -91,7 +105,19 @@ struct HoverButton: View {
         }
     }
 
+    /// Point size the skip glyph is drawn at, matched to the text style the
+    /// symbol would otherwise use so it sits like the icon it replaces.
+    private var glyphPointSize: CGFloat {
+        NSFont.preferredFont(forTextStyle: scale == .large ? .largeTitle : .body).pointSize
```
StudioKeys: `glyphPointSize` calls `NSFont.preferredFont(forTextStyle:)` on every body evaluation. Cache this in `@State` (initialized once) or compute in `init` since the scale is fixed per button instance.

DynamicIsland/components/Notch/NotchHomeView.swift
```diff
@@ -480,7 +479,24 @@ struct MusicControlsView: View {
     private func syncHUDValueIfNeeded(force: Bool) {
         guard shouldShowControlHUDRow else { return }
```
StudioKeys, line 592: Volume HUD animation: `force` path correctly bypasses animation, but `hudValue = target` directly assigns — verify this doesn't conflict with `CustomSlider`'s internal animation state when the slider is being dragged (the `hudDragging` guard should prevent this, but worth a quick manual check).

StudioKeys (changes requested):
## Review Summary

This PR brings excellent polish — Apple-matched sizing, proper skip-glyph animation via `Animatable`, and tighter visual details across the lock screen and notch music panels. The `Animatable` approach for the skip glyph is particularly clever (moving interpolation onto `phase` itself solves the zero-delta problem).

**Blocking**: Validate CHANGELOG check fails — no entry for this PR. Add an entry under `### Changed` (or `### Fixed`) describing the music panel sizing/animation improvements with the PR reference (#742).

Additional actionable items:

1. `SkipTrackGlyph.swift` — Magic numbers, potential trigger coalescing in `.onChange`
2. `LockScreenMusicPanel.swift` — `PanelControlButton` parameter count
3. `HoverButton.swift` — `glyphPointSize` recomputes font on every body evaluation
4. `NotchHomeView.swift` — Volume HUD animation vs slider interaction

DynamicIsland/components/Music/SkipTrackGlyph.swift
vastien: Done in c37925d — `glyphSizeRatio` (0.88), `stepRatio` (0.52) and `travelDuration` (0.26) are now named constants carrying how they were arrived at: rendering this glyph beside `forward.fill` at the same point size and matching triangle size and gap until the resting pair was indistinguishable from the symbol it replaces.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

DynamicIsland/components/Music/SkipTrackGlyph.swift
vastien: Good catch, fixed in c37925d — this was a real hole, not just a theoretical one. Two presses inside one run loop turn do coalesce into a single `.onChange` call with the trigger two higher, and `phase += 1` dropped a skip's worth of travel.

It now advances by the delta (`max(newValue - oldValue, 1)`), so a coalesced double press moves two steps in one animation rather than one. I kept `withAnimation` at the `.onChange` site rather than moving it to the callers: the callers only bump an integer token, and `SkipTrackGlyph` owns its own timing, so pushing the animation outward would spread that across every button that draws one.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

DynamicIsland/components/HoverButton.swift
vastien: Done in c37925d — the two point sizes this button can draw at (`.largeTitle` and `.body`) are resolved once into statics, so `NSFont.preferredFont` is off the body path entirely. Preferred that to `@State` since the value depends only on `scale`, which is fixed per instance and shared across instances of the same scale.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

DynamicIsland/components/LockScreen/LockScreenMusicPanel.swift
vastien: Addressed in c37925d, though not with a config struct — I think there is a better answer to the same complaint.

The count was high because `icon: String` and `skipDirection: Direction?` were two parameters describing one thing, with an implicit rule that the direction wins when both are set. They are now a single `Glyph` value:

```swift
enum Glyph {
    case symbol(String)
    case skipArrows(SkipTrackGlyph.Direction)
}
```

That drops a parameter and, more usefully, makes the invalid combination unrepresentable — a config struct would have preserved it. Call sites read `glyph: .symbol("airplayaudio")` or `glyph: .skipArrows(.forward)`.

Happy to go further into a config struct if you would still prefer one; the remaining seven are genuinely independent (frame, icon size, colour, resting opacity, press interaction, symbol effect, action).

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

DynamicIsland/components/Notch/NotchHomeView.swift
vastien: Checked, no change needed — and you are right that the `hudDragging` guard is what makes it safe. `syncHUDValueIfNeeded(force:)` returns early whenever `hudDragging` is true, so an external volume change cannot animate `hudValue` out from under a drag in progress; the only unanimated assignment is the `force` path, which runs from `onAppear` when the row first appears and should snap to the current level rather than sweep up to it.

Manually verified on the running build: with the notch open, tapping and holding the volume keys glides the fill smoothly, and dragging the slider stays under the cursor with no fighting. That was also the one change in this PR the user confirmed by feel before it was committed.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

vastien:
@StudioKeys all five items have replies, and the CHANGELOG entry is in (the blocking check). I resolved the four that resulted in code changes — constants, trigger coalescing, the cached font metrics, and the parameter count.

I left the volume-HUD thread **open** deliberately: that one asked for a manual check rather than a change, and it seems more useful for you to close it once you agree than for me to mark my own verification as settled. Summary of what I found there: the `hudDragging` guard does prevent the conflict, the only unanimated assignment is the `force` path from `onAppear`, and the glide was confirmed by feel on a running build before it was committed.

One of the five was a genuine bug rather than a nit, worth flagging in case it affects how you want it reviewed: the trigger-coalescing one. Two presses inside a single run loop turn do collapse into one `.onChange` call, and the fixed `phase += 1` silently dropped a skip's worth of travel. It now advances by the delta.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

StudioKeys (approved):

PR #747, opened 2026-08-20
fix(shortcuts): use the documented Cmd+Shift+V for clipboard history

svviitzerland:
Settings has always documented Cmd+Shift+V ("similar to Windows+V on PC"), but
the shortcut was registered as Cmd+Shift+C. It has been `.c` since the first
commit that touched `ShortcutConstants.swift`, so Cmd+Shift+V never worked.

Changing the default alone fixes nothing for existing users:
`KeyboardShortcuts.Name` writes its default into UserDefaults on first launch
and never overwrites it, so the stored Cmd+Shift+C would keep winning. A
one-time migration moves anyone still sitting on the old default.

Custom shortcuts are left alone. The one unavoidable cost is that someone who
deliberately picked Cmd+Shift+C is moved too, because a stored value doesn't
record whether it came from a choice or from the persisted default.

## Summary by CodeRabbit

* **New Features**
  * Updated the default clipboard history shortcut from **Cmd+Shift+C** to **Cmd+Shift+V**.
  * Existing custom shortcuts remain unchanged when they differ from the previous default.

* **Bug Fixes**
  * Added a one-time migration for saved **Cmd+Shift+C** values to **Cmd+Shift+V**.
  * Customizations using the previous default cannot be distinguished from the old default and may also be migrated.


StudioKeys (changes requested):
Changes from codeRabbit can be taken into consideration and fixed if required

svviitzerland:
Hi @StudioKeys, I've applied both suggestions. Could you take another look?

StudioKeys (approved):

PR #748, opened 2026-08-20
feat(clipboard): add Maccy as a clipboard source

svviitzerland:
Adds a "Clipboard Source" setting under Settings › Clipboard, so people who
already use a dedicated clipboard app can hand the feature over to it instead
of running two clipboard histories side by side.

Atoll's own manager stays the default. Picking Maccy:

- sends the notch clipboard icon to Maccy instead of Atoll's panel/popover/tab
- stops Atoll polling the pasteboard
- hides Atoll's own clipboard panel and tabs
- falls back to Atoll's own manager if Maccy isn't installed, so the feature is
  never left dead

Maccy is opened by reopening the app: it answers `applicationShouldHandleReopen`
by toggling its popup, the same thing its status item click does. No
Accessibility permission and no synthetic hotkey.

Atoll also releases its clipboard shortcut while an external app is selected.
When both apps are bound to the same key, one press reaches both: Maccy opened
its popup and Atoll's reopen toggled it straight back shut. That is what the
two defaults did here, and #747 moves Atoll's off Maccy's.

Adding another app is one enum case plus its bundle id, as long as that app also
shows its popup when reopened.

Built on #746: the new picker uses `AppIconImage`, so without that fix the
Maccy icon stretches the same way.

Clipboard Source setting with Maccy selected and detected as running

## Summary by CodeRabbit

* **New Features**
  * Added support for selecting Maccy as an external clipboard source.
  * Added settings to detect, open, refresh, or download supported clipboard providers.
  * Clipboard shortcuts and controls now route to the selected provider automatically.
  * Atoll falls back to its built-in clipboard manager when an external provider is unavailable.

* **Bug Fixes**
  * Fixed picker icons appearing stretched by fitting images to their intended sizes.
  * Improved synchronization when clipboard providers are installed, closed, or changed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,13 +19,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - **Lyrics on the side**: Added ability to show lyrics of the current song when calendar is disabled (#741)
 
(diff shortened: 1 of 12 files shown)


StudioKeys:
@svviitzerland there are fair few comments from CodeRabbit that have not been addressed or looked upon, pls check on those and we could review this up

PR #758, opened 2026-08-21
fix: name the shortcut that actually opens the clipboard history

vastien:
The Clipboard Manager section's description told you to press **Cmd+Shift+V**.

That is not the shortcut. `ShortcutConstants.swift` binds `clipboardHistoryPanel` to `⇧⌘C` by default, and no hardcoded combination could have been right anyway — the shortcut is rebindable under Settings → Shortcuts, so the caption went stale the moment anyone changed it.

## What changed

The footer reads the live binding via `KeyboardShortcuts.Name.clipboardHistoryPanel.shortcut`, renders it with the package's own glyph description (`⇧⌘C`), and refreshes when the binding changes.

`KeyboardShortcuts` posts `shortcutByNameDidChange` when a binding is recorded, but keeps that `Notification.Name` internal to the package, so it is spelled out on the app side to observe it. That is the only way to react to a rebind from outside the package short of vendoring a change into it.

When no shortcut is bound the caption says to assign one rather than naming a combination that does nothing.

## Testing

Built and run on macOS 26.6 (Xcode 26.6). Checked the three states by reading the binding back: default (`⇧⌘C`), rebound, and cleared.

## Summary by CodeRabbit

* **New Features**
  * Clipboard Manager settings now display the currently assigned keyboard shortcut.
  * The displayed shortcut updates automatically when the binding changes.
  * When no shortcut is configured, settings provide guidance to assign one.

* **Documentation**
  * Added a changelog entry describing the updated Clipboard Manager shortcut information.


vastien:
Closing as a duplicate — #743 already replaces this caption with the live binding, via a shared `boundShortcutDescription(for:)` helper that covers the other hardcoded shortcut captions too. That is the better home for it. Sorry for the noise.

PR #772, opened 2026-08-23
Telegram notifications and smart replies (draft, stacked on #771)

vastien:
https://github.com/user-attachments/assets/a977b5e8-59aa-41e8-9106-20232a0ed0b3

> **Draft, and stacked on #771.** It builds on that PR's WhatsApp card rather than beside it, so the diff here includes #771's commits. It should only be looked at once #771 lands, and it still needs work — see *What is not done* at the bottom.

I use Telegram far more than WhatsApp, and #771 looked like it was on its way into the next update, so rather than ask for Telegram there I went ahead and built it. My guess was that nobody was planning it, since it is the more niche of the two — if that guess is wrong, say so and I will drop this.

I also stopped short of two things I would like: **native iOS Messages** and **FaceTime notifications**. Those feel much more like something the project would want to own itself, and I did not want to overstep in case they are already being built. Happy to take either on if they are not.

## Telegram shares the WhatsApp card rather than copying it

The card in #771 — sender, lines, reply box, and the shell that grows the notch around them — is a good piece of work, and none of it is about WhatsApp. So it takes a `ChatService` now and both services share it:

- `SneakContentType.whatsApp(…)` → `.chat(service:…)`
- `WhatsAppNotificationView` → `ChatNotificationView`, `WhatsAppIncomingMessage` → `ChatIncomingMessage`, and so on for the rest of the payload types
- `ChatMessagingProvider` is what the card talks to; `WhatsAppManager` and `TelegramManager` implement it

A service also says what it *can* do. Telegram cannot react or vote from here — its web client puts both behind a hover menu on the message itself, which means opening the conversation and driving someone else's UI for one emoji, and it breaks the moment they move a button. The card does not offer them, rather than offering them and failing.

Nothing about WhatsApp's behaviour changes. The renames are mechanical.

## The engine

Telegram has no local API a third-party app can read a conversation from, and MTProto would mean holding the user's account credentials. So this mirrors #771: the official web client in a window the user never sees.

What it reads is the **chat list**, not the conversation — the one surface Telegram keeps current for every chat at once without any of them being open. That also bounds what a card can say: a photo arrives as the word Telegram writes in the row's preview, because the row is all there is to read.

Three things this cost to get right, each measured rather than guessed.

**A hidden window is a hidden page.** A window parked entirely off screen at zero alpha is occluded, WebKit reports `visibilityState: "hidden"`, and Telegram Web answers by not booting at all. Probed from inside the engine's own view against the same page in a normal browser:

```
offscreen, alpha 0   {"chats":true,"chatsDisplay":"none","authRoot":false,"kids":5,"scripts":1,"vis":"hidden"}
one point on screen  {"chats":true,"chatsDisplay":"",   "chatlist":true, "kids":9,            "vis":"visible"}
```

The static markup loads either way; the client only mounts in the second. Connecting simply never finished. The window now keeps one point of overlap with the screen and a hair of opacity — enough for AppKit to call it visible, too little to see — ignores the mouse, and the page is told it is visible outright as well.

**Reading the row.** A row's title element and the time down its right-hand side both carry the class `row-title`, so the sender read as `_?11:40` until the peer title was asked for by name and the details column excluded. Telegram also writes the typing indicator into the same subtitle the preview comes from, so a card announced "typing" and then added the real message underneath it as a second line of the same conversation.

**Sending.** The text goes in through `insertText` rather than by assigning to the composer: Telegram tracks that `contenteditable` from the input events the browser raises, so an assignment leaves the send button disabled over a box that visibly has words in it. The button is then only clicked once the words are verifiably in the box — clicking send on an *empty* composer does not send nothing, it starts a voice recording. An empty composer afterwards is the confirmation, and anything else is reported as unsent.

**Sign-in** shows a live client rather than a picture of a QR code, since Telegram offers a code *and* a phone number and the second needs somewhere to type. It gets its own view over the shared data store — the session lives in the store, not in either view — and the credentials never pass through this app.

## Settings

Social gains a Smart Replies section at the top — it covers both services — and a Telegram section mirroring WhatsApp's: enable, connection status, connect/disconnect, and a preview notification. The sidebar's BETA badge is now `SettingsTab.isBeta` instead of a chain of `if tab == …` branches each carrying its own copy of the capsule; Social is marked beta, Downloads no longer is.

## Smart Replies

Apple Intelligence writes the suggested replies, on device. `SmartReplyManager` asks `SystemLanguageModel` what it can do and generates three suggestions through guided generation, constrained to differ in *intent* rather than in wording — for a question that usually means yes, no, and a deferral, which is more use than three phrasings of the same answer.

Two things it deliberately does not do. It does not throw at the call site: a refusal, a message the guardrails will not touch, a model that went away mid-session — a notch that cannot suggest anything should simply not offer suggestions, not raise an error over a chat message. And it does not cache availability. Apple Intelligence can finish downloading, or be switched off, while Atoll is running, and a settings pane that asked once at launch would keep saying the wrong thing.

The unavailable reasons are kept apart rather than collapsed to a bool, because they want different words: "this Mac cannot" and "the model is still downloading" are different news, and only one of them is worth waiting for. The toggle disables itself and says which.

It is **off by default**. It reads message text to do its job, and that is not something anyone should discover after the fact — the section says plainly that the text stays on the machine.

Deployment target is macOS 14.6, so the whole path is behind `canImport(FoundationModels)` and `@available(macOS 26, *)`, and compiles to a stub that returns no suggestions on anything older.

## What is not done

Being honest about the state of this, which is why it is a draft:

- **It does not link on this branch.** #771 ships a resolved-versions bump whose newer Lottie drops the initialiser `LottieUI` calls. That is not this branch's to fix, so it is left alone — locally I build with dev's `Package.resolved`.
- Cards carry the chat list's one-line preview, not the message body. Media shows up as Telegram's own wording for it.
- No reactions, no poll voting, no attachments.
- The selectors come from the Telegram Web K source and from probing the live page. They will break when Telegram reworks its DOM, the way any scraper does.
- The card's layout wants work at Telegram's proportions — the reply row can overflow.
- **Smart Replies is unverified against a live model.** Apple Intelligence reports `modelNotReady` on my machine and its generative-models asset directory is empty, so the availability path and the disabled-with-a-reason UI are what has actually been exercised. The generation call itself is written against the framework's guided-generation API and has not yet produced a real suggestion.


vastien:
07a8b92 fixes the notch jerking downward and briefly showing the desktop through the top of the screen whenever a card grew — a second message arriving, a reply row changing height.

`setFrame(_:display:animate:)` runs its animation on the main thread and blocks it, so SwiftUI draws no frames at all while the window grows. AppKit fills them with the layer's existing contents, anchored by default to the bottom left — the wrong end for a window hanging from the top of the screen.

A growing window is no longer animated. Growing it in one step costs nothing visually: the window is transparent and the card inside does its own animating, so only the card is ever seen. Shrinking still animates, so a card is never cut off mid-animation.

The hosting view also anchors layer contents to the top left. That alone did **not** fix it — I tried it first and the jerk stayed — but it's still the correct anchor for the animated shrink path that remains.

This belongs here rather than in its own PR: the top-aligned animated resize it modifies only exists on this branch, so there was nothing to patch on `dev`. Verified on a local build.

PR #784, opened 2026-08-26
fix: keep Lock Screen Glass in one place

vastien:
Settings had two **Lock Screen Glass** sections — one on Appearance, one on Lock Screen — driving the same defaults.

The Appearance copy had the same `Material` picker, the same glass-mode picker, and variant sliders for the music panel and timer widget. Those sliders already exist on the Lock Screen tab; the Appearance copy's own caption said so ("Changes mirror the Lock Screen tab"), and its footer told you to "Configure lock screen materials from the Appearance tab" — which is the tab you were already on.

So the duplicate is gone and the Lock Screen tab keeps it, on the grounds that it is where the widgets being described are configured. Also removes the four bindings and the variant range in `Appearance` that only fed the removed section.

No defaults changed and no keys renamed, so nobody's existing configuration moves — only where you go to change it. Settings search had no entries pointing at the removed rows.

## Summary by CodeRabbit

- **Improvements**
  - Moved the live glass preview from the duplicate **Appearance** settings to the **Lock Screen** tab.
  - Consolidated Lock Screen glass controls in one location for a clearer settings experience.
  - Added a liquid-glass preview to the lock-screen music variant slider.
  - Removed duplicate lock-screen music, timer, and glass controls from **Appearance**.
  - Timer-specific glass settings remain available in the timer settings.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,7 +28,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **Lyrics on the side**: Added ability to show lyrics of the current song when calendar is disabled (#741)
 
 ### Changed
-- **Lock screen glass has one home**: the Lock Screen Glass section is now only on the Lock Screen tab. Appearance carried a second copy of the same material and glass-mode pickers, with its own variant sliders whose caption admitted they mirrored the other tab — two places to change one setting, and neither obviously the real one.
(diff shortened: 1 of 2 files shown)


vastien:
Folded into #787, which already contains both commits from this branch (0c6500c and f24d734) and is based on `dev` — so its diff carries this change in full, no separate review needed.

Closing to keep it to one PR: it's all the same job, making the settings panel navigable, and there's no reason to make you read two.

PR #796, opened 2026-08-28
feat: favourite the current song in Cider

vastien:
Cider already worked as a media source, but only for playback. It is followed through macOS Now Playing, which carries no favourite command, so the Favorite Song control had nothing to call and sat inert when Cider was the source.

Cider exposes a local HTTP API for external applications, and it answers both halves of the question — `GET /api/v1/playback/now-playing` reports `inFavorites`, and `POST /api/v1/playback/set-rating` writes one back. So unlike TIDAL, where the heart has to be a read-only indicator because the app accepts no write at all, Cider's heart is live.

### Notes on the choices

- **Un-favouriting sends rating `0`, not `-1`.** A dislike is a separate opinion in Apple Music and would follow the user around their recommendations; that is not what clearing a favourite should mean.
- **The token goes in an `apptoken` header, bare, with no `Bearer` prefix.** Cider's own RPC documentation names that header `apitoken`, which its server does not accept. There is a test pinning this, because it is the single detail most likely to regress.
- **Every failure reads as unknown, not as "not favourited".** Cider closed, its API switched off, a rejected token, nothing playing — all return `nil`, so the heart never shows a confident wrong answer.
- **`canEverFavorite` stays true regardless**, so the control keeps its place in the palette while Cider is closed, per the capability model.
- Playback is untouched. The token is only needed for favouriting, and the settings footer says so, so it does not read as a requirement for using Cider as a source at all.

### Setup, for anyone trying it

In Cider: **Settings → Connectivity → Manage External Application Access**, switch the API on, and paste the generated token into Atoll's media settings. If you turn Cider's authentication off instead, leave the field empty and no header is sent.

## Testing

**I do not have Cider installed, so I could not verify this against a running instance.** Nothing here has been exercised end to end against the real app, and that is the main thing a reviewer should weigh. Anyone with Cider who can confirm the heart reflects and changes the real favourite state would be doing me a favour.

What *is* verified:

- 12 unit tests covering the wire format against the JSON shapes in Cider's RPC documentation — all passing, along with the rest of the suite. They pin the two request paths, the `apptoken` header (including that no header is sent when no token is set, and that `apitoken`/`Authorization` are never used), the `{"rating": 1}` / `{"rating": 0}` bodies, and that a 401, a 500, a connection failure, and a missing field each read as unknown rather than as a definite answer.
- Builds clean.

The endpoints and their payloads come from Cider's published RPC documentation rather than from observation, so if Cider v2 has since changed either, the tests would still pass while the feature did not work. That is the specific risk of not having the app.

### Stacked on #788

This needs the favouriting capability model (`canEverFavorite` / `supportsFavoriting` / `favoritingIsReadOnly`) from #788, which is not in `dev` yet, so this branches from it and should merge after. The diff shown here against `dev` will include #788's commits until that lands.

## Summary by CodeRabbit

- **New Features**
  - Added Favorite Song support for Apple Music, Spotify, YouTube Music, and Cider.
  - Added Cider API token configuration in Settings.
  - Added TIDAL shuffle and repeat controls through its Playback menu.
  - Favorite controls now appear across supported sources and are disabled when unavailable or read-only.
  - Updated song identity handling for YouTube Music favorites.

- **Bug Fixes**
  - TIDAL playback modes now refresh and toggle correctly.
  - Favorite state and unsupported-source behavior are reported more accurately.


vastien:
Folded into #788 — the Cider favouriting commit now sits on `feat/favorite-current-song`, so it ships with the rest of the media source control work rather than as a separate PR against the same capability model it depends on.

Closing this one; nothing is lost. Same commit (`3c70c65`), same tests.

PR #806, opened 2026-09-01
Pin the current lyric under the closed notch

vastien:
Lyrics currently live inside the open notch, which means they are visible exactly while you are hovering — and you cannot hover the notch and do anything else at the same time. This keeps the current line under the notch after it closes, so the lyric is readable while you work.

**Off by default** (`pinLyricsWhenClosed`), and it does nothing unless lyrics are already enabled.

## Preview

The current lyric pinned under the closed notch

The notch is closed and unhovered here — artwork, spectrum, and the current line underneath, with the sweep highlight tracking the vocal.

## The pin lives with the lyrics

A pin button in the Lyrics panel header, filled while pinned. It is deliberately not a Settings-only toggle: the thing it controls appears when the notch *closes*, so the control belongs on the panel that has the lyrics in it, next to what it affects.

The button is filled while active rather than only tinted, because the strip it controls is off screen whenever the button is on screen — the two are never visible together, so the button has to carry the state on its own.

## Why this was harder than it looks

Three attempts before this one, all of which deformed the notch.

`NotchLayout()` is a `VStack(alignment: .leading)` and **the panel sizes itself to its widest child**. Anything added to that stack that reports a width stretches the panel into a full-width slab, which re-centres the music row and slides the spectrum under the physical notch cutout. `maxWidth: .infinity` does it. A plain sized `Text` does it. A `GeometryReader` does it worst of all, because a `GeometryReader` returns whatever width it is *offered* — which is the whole screen — rather than what its content needs.

The primitive that works is **padding plus an overlay**:

```swift
content
    .padding(.bottom, isShowing ? Self.stripHeight + Self.stripGap : 0)
    .overlay(alignment: .bottom) { lyricLine }
```

Padding can only add height; it cannot widen the panel. An overlay is proposed the parent's size and never changes it. So the strip gets its own row without the layout ever asking it how wide it would like to be.

It ships as a `ViewModifier` (`.pinnedLyrics(isVisible:)`) rather than a view for the same reason — a view would have to be *placed* somewhere in that stack, and every place in that stack is the wrong place.

## Animation

The strip fades and slides in over 0.28s when a line first becomes available, and line-to-line changes cross-fade at 0.18s, so the strip does not pop into existence when the song reaches its first lyric. The existing `lyricSweep` highlight carries over unchanged, driven by `TimelineView(.animation(paused: !isPlaying))` so a paused track is not animating for nothing.

An earlier commit scrolled long lines as a marquee. It looked worse in practice and changed how the lyrics read, so it is reverted in this branch (c68d9b9) — the line truncates instead.

## Tests

`PinnedLyricsTests` covers `PinnedLyricsView.hasCurrentLine(index:lineCount:)`, the index-bounds check that decides whether there is a line to draw at all — including the -1 "before the first lyric" case that the lyrics engine emits, which is what put an empty strip on screen during development.

## Summary by CodeRabbit

* **New Features**
  * Added an option to keep the current lyric line visible beneath the closed notch.
  * Added a pin control in the Lyrics panel and a corresponding Media setting.
  * Pinned lyrics follow playback, animate smoothly, and truncate long lines without widening the notch.
  * The feature is off by default and appears only when supported lyric and playback conditions are met.
  * Improved positioning to avoid covering frontmost app menus.
* **Tests**
  * Added coverage for lyric-line availability and index boundaries.


vastien:
image
image

StudioKeys (changes requested):
Left detailed review below

StudioKeys (changes requested):
## Changes Required (CHANGES_REQUESTED)

### 1. Reverted Marquee Experiment
- **Commit**: `c68d9b9` reverts `26810f0` — "feat: scroll the pinned lyric instead of truncating it"
- **Note**: Author reverted after "looked worse in practice"; truncation is correct for single-line strip
- **Action**: Add comment in code why marquee was reverted (for future maintainers)

### 2. `stripHeight` / `stripGap` Magic Numbers
- **Lines 37-38 in `PinnedLyricsView.swift`**:
```swift
private static let stripHeight: CGFloat = 13
private static let stripGap: CGFloat = 4
```
- Could use `.caption` font metrics or derive from `stripHeight` + line spacing

### 3. Tint Brightness Floor
- **Line 67**: `ensureMinimumBrightness(factor: 0.6)` — ensure this helper handles all color spaces

### 4. Sneak Peek Exclusion
- **ContentView.swift line 2400**: `&& !isSneakPeekVisibleOnCurrentScreen`
- **Note**: Correct — both draw in same strip below notch; volume HUD would jitter otherwise
- Consider if this should also exclude other transient overlays

### 5. Animation Durations Hardcoded
- **Lines 63-64**: `0.28` and `0.18` — consider `Defaults` keys if users want tuning

### 6. No Width Constraint on Overlay
- Overlay capped by parent; could add `.frame(maxWidth: closedContentWidth - 20)` for explicit cap

---

## Layout Problem (for context)

The PR documents 3 failed attempts before the working overlay solution:

1. **Width Capping** (`6e76b382`) — only shrank the slab
2. **GeometryReader** (`aa6c88ff`) — returns offered width (whole screen)
3. **maxWidth: .infinity** — panel widens, music row re-centers, spectrum under notch

**Working Solution**: Padding + Overlay as ViewModifier (`ac8314d3`)

---

vastien:
Thanks — went through all six. Three are addressed in `09fad56`, three I'd push back on; reasoning below so you can overrule any of it.

**1. Marquee revert undocumented — done.** There was no trace of the experiment in the code, so the next person to see a truncated lyric would have had the same idea and tried it again. The comment now says it was tried, that it just didn't look up to par next to the rest of the UI in testing, and that truncation is the easy way out rather than the considered answer — so anyone picking it back up starts from how it ought to look rather than from this.

**2. `stripHeight` / `stripGap` magic numbers — done for the one that was actually a metric.** `stripHeight` was written as `13` next to a line set at `10pt`, which is only correct for as long as nobody touches the font. Both now come from a single `lyricFontSize`, with the height derived as `(ascender - descender + leading).rounded(.up) + 1` — one line of that exact font, plus a point so a descender isn't shaved by the frame. That evaluates to `13`, so nothing moves on screen; it just can't drift now.

`stripGap` I left as `4` with a comment. It's the space between the strip and the notch body — a gap judged by eye, not derived from the text, and deriving it from `stripHeight` would tie it to something it has no relationship to.

**3. `ensureMinimumBrightness` colour spaces — checked, it's fine.** It calls `usingColorSpace(.sRGB)` and guards the result, returning the original colour if conversion fails, so `getRed(_:green:blue:alpha:)` is never reached on a non-RGB `NSColor`. No crash path there.

While in it I did find a separate defect, pre-existing and not from this PR: `let scale = factor / perceivedBrightness` divides by zero for a pure-black input, and `min(0 * .infinity, 1.0)` is `NaN`, so every component comes out `NaN`. It needs artwork whose average colour is exactly black — reachable, if rare. It's shared code in `NSImage+Extensions.swift` with other callers, so I'm putting the guard on its own branch rather than smuggling an unrelated fix into this one.

---

**4. Sneak-peek exclusion — no change.** You marked the existing code correct, and I agree. Broadening it to "other transient overlays" without a specific one that conflicts would be guessing at a list; the moment a real one collides it should be added by name.

**5. Animation durations as `Defaults` keys — no change.** `0.28` and `0.18` are interaction tuning, not preference. They're already commented with *why* they differ (the strip arriving is a bigger event than one line replacing another, and sharing a duration made every line change feel like the panel resizing). Exposing them adds two settings nobody asked for and two ways to make the animation feel wrong.

**6. Explicit `maxWidth` cap — no change.** The parent already bounds the overlay; a second cap is a redundant constraint that can drift out of step with the first and start clipping when the notch width changes.

Also merged current `dev` (`4782b45`) — the PR was conflicting on `CHANGELOG.md` and `project.pbxproj` after #805 landed. 222 tests pass.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

StudioKeys (approved):

PR #809, opened 2026-09-02
fix: show untimed lyrics as lines, and prefer lyrics that have timings

vastien:
Two related lyrics problems, both visible on the same tracks.

### Untimed lyrics were one enormous line

When a track only has plain lyrics available, the whole body was kept as a single `LyricLine`. The panel then treated the entire song as the line currently being sung — sweeping the highlight across all of it at once, straight through the middle of words.

Plain lyrics are now split on newlines into their own lines and marked `isTimed: false`. The sweep is skipped for them and the sync timer never starts, so they are simply readable and nothing pretends to know where in them you are.

### Timed lyrics lost the ranking to untimed ones

Timings were worth 3 points in the match ranking, against 20 available for title/artist/album agreement. A result with no timings but an exactly matching album could therefore beat one that had timings and agreed slightly less precisely — and the track showed as plain text with no line ever lit, even though timed lyrics existed for it.

Whether the words can be followed at all is now decided first, and the existing score only separates results that are alike in that.

### Tests

11 new tests across `UnsyncedLyricsTests` and `SyncedLyricsPreferenceTests`; 180 pass locally.

## Summary by CodeRabbit

- **Bug Fixes**
  - Timed lyrics now take priority over untimed lyrics when selecting matches.
  - Plain lyrics are displayed as separate, cleaned-up lines.
  - Untimed lyrics no longer trigger synchronization, preventing incorrect highlighting and scrolling.
  - Lyrics now fall back to NetEase Cloud Music when no suitable lyrics are found.

- **Documentation**
  - Added changelog entries covering usage and cost tracking, pricing updates, countdown parsing, lyrics fallback, and recent interface improvements.


StudioKeys:
@vastien there is an unresolved comment from coderabbit, look into it please

StudioKeys (approved):
Review complete. All CI passing, merge state CLEAN.

## Summary
Two tightly related lyrics bugs fixed cleanly:

1. **Untimed lyrics as lines** —  splits plain body on newlines, trims, filters empty, marks . Highlight sweep skipped, sync loop never starts,  stays .

2. **Timed lyrics preferred** —  now primary sort key via tuple . Exact album match (4 pts) no longer beats timed result (was 3 pts). Score only tiebreaks among same-timed results.

## Code Quality
- Pure  — testable, handles whitespace/blank lines
- Tuple ranking  — clean priority, single source of truth
- Early returns for untimed — no false sweep, no sync timer
-  computed property — single check used in 3 spots
- 11 comprehensive tests covering exact bug scenarios

## Minor Note
 init default  preserves existing callers; new static method for untimed. No UI changes needed.

**Ready to merge.**

PR #831, opened 2026-09-11
fix: stop recurring Claude Keychain prompts after credential refresh (#820 follow-up)

MoFeng2223:
## Problem and relationship to #820

This follows up on my earlier PR #820. That change fixed a real credential-persistence problem: after refreshing Claude's OAuth tokens, Atoll now saves the rotated pair back to its original source, so Claude Code does not keep using an invalidated pair and lose its login.

However, #820 did not preserve compatible Keychain access across repeated refreshes. On my Mac, choosing **Always Allow** was only temporarily effective: another credential update was followed by another Keychain authorization prompt when Claude accessed the item.

Atoll wrote the secret through native `SecItemUpdate`, while Claude Code used `/usr/bin/security`. Updating a file-based Keychain secret can recreate its partition ACL under the writer's identity. During the reported failures, the item's partition list contained Atoll's code identity but no longer contained `apple-tool:`; subsequent `security` reads logged `ACL partition mismatch: client apple-tool:` and prompted again. This explains the remaining permission problem without undoing #820's necessary token write-back.

## Validation after continued use

I installed this fix on my Mac on **September 8, 2026**, and tested it during normal use through **September 11**. The repeated Keychain authorization prompts have not appeared again during that time.

The initial fix passed a local Release build and code-signature verification before I installed it. Both macOS CI builds passed for that revision. The follow-up changes below add protection for a closed input pipe and a tracked test runner; they have passed local compilation and regression tests.

A standalone regression test was rerun after rebasing and passed. It uses disposable, UUID-named Keychain items and performs three cycles of Atoll-style writes, Claude-style writes, and reads. It checks that `apple-tool:` and the partition set remain unchanged, payloads survive exactly (including Unicode and escaping), and other accounts remain untouched. It also covers missing accounts and oversized or invalid input. Test items are removed afterward; real Claude credentials are not used by the test.

The added closed-pipe regression runs in a subprocess with default `SIGPIPE` handling. With the reader already closed, the write reports failure and the subprocess exits normally. The tracked runner checks for macOS, and the test requires an unlocked, readable and writable default login Keychain before creating any disposable credentials.

## Implementation

- Route both Claude credential reads and write-backs through `/usr/bin/security`, matching Claude Code's access identity and avoiding alternating native-app/tool partition identities. Keep #820's write-back to the original source and preservation of credential JSON metadata.
- Scope operations to the selected service and account, including hash-suffixed Claude items. Check that the original item exists before the tool's update operation.
- Send updated secrets through the tool's standard input, never through a shell or process arguments. Escape its interactive syntax, decode its hexadecimal output when needed, and suppress potentially sensitive error output.
- Set `F_SETNOSIGPIPE` on the input pipe's writing descriptor and report write errors as operation failures. Close the parent's unused reading descriptor after launch so an early child exit is observable. This does not change the application's global signal handling.
- Respect the tool's 4095-byte command limit. Reject invalid or oversized commands and check available space before consuming a shared refresh token; leave oversized records for Claude Code to refresh. The existence check followed by `-U` is not an atomic update-only operation, so a concurrent deletion remains a limitation.

This PR addresses recurring Keychain authorization after credential rotation. The separate quota-display case where the API returns `five_hour.resets_at: null` is unchanged.

Regression test:

```sh
./scripts/test_claude_keychain.sh
```

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed repeated macOS Keychain prompts after Claude token rotation.
  - Improved Claude credential refresh handling by keeping credential reads and updates consistent.
  - Prevented oversized credential records from being truncated; Claude Code can refresh these records safely instead.
  - Improved reliability when updating Claude credentials, including safer handling of special characters and interrupted system operations.
  - Added safeguards to preserve credential integrity when Keychain updates cannot be completed safely.


StudioKeys (changes requested):
## Review: fix: stop recurring Claude Keychain prompts after credential refresh (#820 follow-up)

This is a well-targeted fix for a real issue: the partition ACL mismatch between Atoll's native `SecItemUpdate` writes and Claude Code's `/usr/bin/security` reads causing repeated Keychain prompts after token rotation.

### Key Strengths
- Root cause correctly identified: alternating native/tool partition identities reset `apple-tool:` ACL entry
- Solution uses same `/usr/bin/security` identity for both reads and writes via `ClaudeKeychainStore`
- Properly scopes operations to service + account (not just service) matching Claude Code
- Secrets passed via stdin (not shell/args) with proper escaping
- SIGPIPE suppression via `F_SETNOSIGPIPE` on write descriptor
- Command length guard (4095 bytes) prevents silent truncation
- Oversized records left for Claude Code to refresh, preserving login
- Comprehensive regression test covering 3 Atoll/Claude cycles, partition survival, exact payload round-trip, account isolation, missing/oversized/invalid input

### Required Changes

**ClaudeQuotaClient.swift** - Replace `print()` statements with project's `Logger` utility (used throughout codebase in `DynamicIsland/utils/Logger.swift`). Two occurrences:
- Line ~196: file write failure
- Line ~201: Keychain write failure

### Notes
- Test at `tests/test_claude_keychain.swift` is a standalone executable (has `@main`), not an XCTest in `DynamicIslandTests/`. This is acceptable for a regression test requiring Keychain access, but note it won't run in standard `xcodebuild test`.
- Test requires unlocked default login Keychain — documented in script but may limit CI execution (acceptable for local validation).
- CHANGELOG entry added — good.

MoFeng2223:
@coderabbitai I checked both findings against the current code and tested the account-attribute behavior on my Mac.

**1. Command length: the current check already counts bytes.**

`ClaudeKeychainStore.updateCommand(...)` returns `Data?`, not `String?`. It constructs that data from the command's UTF-8 representation and checks `data.count <= maximumCommandBytes`. The caller's unwrapped `command` is therefore `Data`, so `command.count + 512` is also a byte-count check. The two checks use the same units, including for non-ASCII input. No change to `.utf8.count` is needed for this finding.

**2. Missing account: reproducible as a Keychain edge case, but not observed from the normal writers I checked.**

On macOS 26.5.1, I created disposable generic-password items with `SecItemAdd`, both omitting `kSecAttrAccount` and setting it to an empty string. Both succeeded, and attribute queries returned no account key. An explicit empty account through `/usr/bin/security` produced the same result. With a newer such item and an older account-bearing item, I reproduced `freshestGenericPassword` selecting the newer item and returning nil. All temporary items were deleted afterward; no real Claude credential was modified.

I also inspected the installed Claude Code 2.1.263 implementation. Its Keychain writer explicitly passes `-a` in both write paths. The account helper uses the system username and falls back to `claude-code-user`; its validation rejects an empty name. This PR's Atoll write-back preserves the account from the selected existing item and rejects nil accounts instead of removing the account attribute. My actual Claude Keychain item has a nonempty account.

Thus I do not currently have evidence that normal writes by this Claude Code version and Atoll generate account-less items. The reproduced case is a valid defensive-hardening concern for an item created through another path, rather than a confirmed failure of those normal writers. Apple's DTS explanation also confirms that empty attributes may be omitted from returned dictionaries, while noting implementation differences on macOS: https://developer.apple.com/forums/thread/822626.

StudioKeys (approved):

